PR #2, opened 2022-02-22
Add application extensions.

jortel:
Add Application.Extensions to support application _type_ specific properties.
Needed for MVN coordinates for java application binaries.
Also, suggested that it would be useful for plugins to augment the application with information not suitable for buckets.


mansam (approved):
LGTM

PR #3, opened 2022-02-24
GET/POST with value as body

jortel:
Small change to support GET and PUT body begin the setting value.

Example:
```
curl -s -X PUT localhost:8080/settings/git.insecure.enabled -d "true"
curl -s -X GET localhost:8080/settings/git.insecure.enabled
true
```

The POST (create) body will continue to be `api.Setting`.
The GET (list) result will continue to be `[]api.Setting`.


mansam (approved):
LGTM

PR #4, opened 2022-02-24
Add method to find identity for application and kind.

jortel:
Add method to find identity for application and kind.
Common for addons.

### DIFF
--- a/addon/application.go
+++ b/addon/application.go
@@ -41,3 +41,28 @@ func (h *Application) Update(r *api.Application) (err error) {
 	}
 	return
 }
+
+//
+// FindIdentity by kind.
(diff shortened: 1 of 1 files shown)


addon/application.go
```diff
@@ -41,3 +41,28 @@ func (h *Application) Update(r *api.Application) (err error) {
...
+	err = h.client.Get(path, &list)
+	if err != nil {
+		return
+	}
+	for i := range list {
+		r = &list[i]
+		if r.Kind == kind {
+			m := r.Model()
+			err = m.Decrypt(Addon.secret.Hub.Encryption.Passphrase)
+			r.With(m)
```
mansam, line 59: nitpick: `r.With(m)` should probably be after the error conditional.

mansam (approved):
LGTM, one tiny nitpick

PR #6, opened 2022-02-24
addon: use workingDir as home.

jortel:
addon: use workingDir as home.
The workingDir will be a mounted volume.  We don't want to depend on the elasticity of the _tmpfs_.
This also needs to be the _home_ directory because some CLI tools write files to $HOME by default.


mansam (approved):
LGTM

PR #7, opened 2022-02-25
Fix identity updates.

jortel:
When using db.Updates(), the hooks are not fired which in this case is really bad since the BeforeSave() ensures fields are encrypted.

Also revert part of https://github.com/konveyor/tackle2-hub/pull/6.  Somehow, the changed $HOME was not respected by maven.  Found a simpler approach.

Add `Handler.fields()` method using to build a _map_ of a model's fields.  Considered adding `BaseModel.Map()` but since go does not support polymorphism, The receiver IsA BaseModel not the concrete type.

Refit the TaskHandler.Update() to use fields().

### DIFF
--- a/api/identity.go
+++ b/api/identity.go
@@ -148,10 +148,9 @@ func (h IdentityHandler) Update(ctx *gin.Context) {
 		return
 	}
 	m := r.Model()
-	db := h.DB.Model(&Identity{})
+	db := h.DB.Model(m)
 	db = db.Where("id", id)
-	db = db.Omit("id")
-	result := db.Updates(m)
+	result := db.Updates(h.fields(m))
 	if result.Error != nil {
 		h.updateFailed(ctx, result.Error)
 		return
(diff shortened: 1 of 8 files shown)


api/identity.go
```diff
@@ -148,10 +148,9 @@ func (h IdentityHandler) Update(ctx *gin.Context) {
 		return
 	}
 	m := r.Model()
-	db := h.DB.Model(&Identity{})
+	db := h.DB.Model(m)
```
jortel, line 151: Must use a real model instance here for the _hooks_ to fire.

model/base.go
```diff
@@ -7,8 +7,8 @@ import (
 //
 // Model Base model.
 type Model struct {
-	ID         uint `gorm:"primaryKey"`
-	CreateUser string
+	ID         uint      `gorm:"<-:create,primaryKey"`
+	CreateTime time.Time `gorm:"<-:create,autoCreateTime"`
+	CreateUser string    `gorm:"<-:create"`
```
jortel, line 12: Add `<-:great` to make the PK and other fields _read only_.

mansam (approved):
LGTM

PR #9, opened 2022-02-25
Addon volumes

jortel:
Fix custom addon volumes not mounted in the container.

### DIFF
--- a/addon/adapter.go
+++ b/addon/adapter.go
@@ -105,12 +105,6 @@ func (h *Adapter) Client() *Client {
 // newAdapter builds a new Addon Adapter object.
 func newAdapter() (adapter *Adapter) {
 	//
-	// Working directory.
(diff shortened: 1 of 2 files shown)


addon/adapter.go
```diff
@@ -104,12 +104,6 @@ func (h *Adapter) Client() *Client {
 //
 // newAdapter builds a new Addon Adapter object.
 func newAdapter() (adapter *Adapter) {
-	//
-	// Working directory.
-	wDir := Settings.Path.WorkingDir
-	_ = os.MkdirAll(wDir, 0755)
-	_ = os.Setenv("HOME", wDir)
-	_ = os.Chdir(wDir)
```
jortel, line 112: Reverting.  maven did not respect it so found a better solution.

mansam (approved):
LGTM

PR #11, opened 2022-03-02
Add Proxy: enabled and excluded hosts list.

jortel:
Add Proxy: enabled and excluded hosts list.

Also applies new pattern to updates.  Needed to support enabled=0.


mansam (approved):
LGTM

PR #12, opened 2022-03-02
Implement Keycloak-based auth flow

mansam:


settings/auth.go
```diff
@@ -0,0 +1,54 @@
...
+		ClientID     string
+		ClientSecret string
+	}
+	// Addon API access secret
+	AddonAccessSecret string
+}
+
+func (r *Auth) Load() (err error) {
+	var found bool
+	r.Required = getEnvBool(EnvAuthRequired, true)
```
jortel, line 32:
```
if !r.Required {
  return
}
```
seems simpler.

settings/auth.go
```diff
@@ -0,0 +1,54 @@
...
+	Required bool
+	// Keycloak client config
+	Keycloak struct {
+		Host         string
+		Realm        string
+		ClientID     string
+		ClientSecret string
+	}
+	// Addon API access secret
+	AddonAccessSecret string
```
jortel, line 27: I assume you're using the name _AccessSecret_ to differentiate from k8s secrets.  Perhaps instead of _AccessSecret_ we could consider `AddonToken`.

settings/auth.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func (r *Auth) Load() (err error) {
+	var found bool
+	r.Required = getEnvBool(EnvAuthRequired, true)
+	r.Keycloak.Host, found = os.LookupEnv(EnvKeycloakHost)
+	if r.Required && !found {
+		panic("KEYCLOAK_HOST required")
```
jortel, line 35: On one hand, I like the panic.
But, we don't do this anywhere else but perhaps there are no sane defaults.  Thinking about running the hub locally.

api/addon.go
```diff
@@ -25,9 +26,11 @@ type AddonHandler struct {
 //
 // AddRoutes adds routes.
 func (h AddonHandler) AddRoutes(e *gin.Engine) {
-	e.GET(AddonsRoot, h.List)
-	e.GET(AddonsRoot+"/", h.List)
-	e.GET(AddonRoot, h.Get)
+	routeGroup := e.Group("/")
```
jortel, line 29: :+1: routeGroup

auth/provider.go
```diff
@@ -0,0 +1,129 @@
...
+	for _, s := range claimScopes {
+		scope := NewKeycloakScope(s)
+		scopes = append(scopes, &scope)
+	}
+	return
+}
+
+//
+// NewKeycloakScope builds a Scope object from a string.
+func NewKeycloakScope(s string) (scope KeycloakScope) {
```
jortel, line 106: Seems to only be used by the _KeyCloak_ object.
Consider `KeyCloak.newScope()`

settings/auth.go
mansam: good call, must have shut my brain off while writing that section. ;)

settings/auth.go
mansam: Heh, I pushed that exact change immediately before seeing your review.

settings/auth.go
mansam: Sure, I can put together a sample realm with some defaults.

PR #14, opened 2022-03-08
Fix SEGV setting auth header in NIL Header object.

jortel:
Fix SEGV setting auth header in NIL Header object.
Use _const_ for header key for better coupling.


mansam (approved):
LGTM

PR #16, opened 2022-03-08
Add missing FKs on Proxy and Bucket.

jortel:
Add missing FKs on Proxy and Bucket.  Added additional (relation) fields apparently needed by GORM to generate the FK constraints.

Add _Resource_ methods:
- ref(uint, Named)
- refPtr(*uint, Named)
- idPtr(*Ref)


mansam (changes requested):
LGTM

mansam (approved):

PR #19, opened 2022-03-10
Application and Task IsA bucket.

jortel:
Problem and solution discussed [here](https://github.com/konveyor/tackle2-hub/issues/20).


api/taskgroup.go
```diff
@@ -0,0 +1,363 @@
...
+	err := ctx.BindJSON(r)
+	if err != nil {
+		return
+	}
+	g := r.Model()
+	g.ID = id
+	err = h.DB.Transaction(
+		func(tx *gorm.DB) (err error) {
+			db := tx.Model(g)
+			db = db.Omit("Tasks")
```
mansam, line 147: You can use `clause.Associations` in a call to Omit() rather than having to name each association. We can also probably use that to preload associations in most cases.

mansam (changes requested):
One small fix needed but otherwise LGTM

api/taskgroup.go
jortel: So db.Omit(clauses.Associations)?

mansam: Yep, `db = db.Omit(clause.Associations)`

mansam: GoLand should automatically figure out to import `gorm.io/gorm/clause`

mansam (approved):
LGTM

PR #23, opened 2022-03-12
Add BaseHandler methods: pk() and fix fields().

jortel:
The pk() is a convenience for getting the `ID` parameter as an integer. This will be useful for update methods that need to coerse the model.ID which is needed for Association methods and eliminates the need to set using _Where_.


mansam (approved):
LGTM

PR #37, opened 2022-03-18
Tasking updates

jortel:
Task/Group status renamed: `state`.

Task/Group submit support PUT with/without body.  This is accomplished by _chaining_ the submit and update handlers.  The submit handler modifies the body in the request or injects (fetched from DB) ensuring the `state=Ready`.
  
TaskGroup create list of task in json field instead of creating the Tasks immediately.  On submit, the group properties are propagated and tasks created.


mansam (approved):
LGTM

PR #47, opened 2022-03-29
Fix identity update.

jortel:
Encrypt() needs to be called before h.fields() because the hooks are called by db.Updates().

Simplified the idempotent behavior of both Encrypt() and Decrypt() by moving the clearing of the encrypted fields to the hooks.
Explicitly call Encrypt() in the handler Create() and Update() leaving the hooks to perform an Encrypt() as a safety net.  This will result in the Encrypt() being called twice but this is harmless.

### DIFF
--- a/api/identity.go
+++ b/api/identity.go
@@ -151,6 +151,11 @@ func (h IdentityHandler) Update(ctx *gin.Context) {
 	m := r.Model()
 	m.ID = id
 	db := h.DB.Model(m)
+	err = m.BeforeUpdate(db)
+	if err != nil {
+		h.updateFailed(ctx, err)
+		return
+	}
 	result := db.Updates(h.fields(m))
 	if result.Error != nil {
 		h.updateFailed(ctx, result.Error)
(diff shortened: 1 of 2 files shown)


api/identity.go
```diff
@@ -210,10 +215,6 @@ func (r *Identity) With(m *model.Identity) {
 	r.Kind = m.Kind
 	r.Name = m.Name
 	r.Description = m.Description
-	r.User = m.User
-	r.Password = m.Password
-	r.Key = m.Key
-	r.Settings = m.Settings
```
jortel, line 216: Just to be safe, these fields never need to be returned by the API.

api/identity.go
```diff
@@ -228,7 +229,6 @@ func (r *Identity) Model() (m *model.Identity) {
 		Password:    r.Password,
 		Key:         r.Key,
 		Settings:    r.Settings,
-		Encrypted:   r.Encrypted,
```
jortel, line 231: The `encrypted` is never relevant when passed to the API.

PR #50, opened 2022-03-31
Add AdoptionPlan handler

mansam:
Generates an app dependency graph and sorts it into topological
order, assigning weights based on the effort of each app's
most expensive chain of dependencies.

### DIFF
--- a/api/adoptionplan.go
+++ b/api/adoptionplan.go
@@ -1,0 +1,238 @@
+package api
+
+import (
+	"github.com/gin-gonic/gin"
+	"github.com/konveyor/tackle2-hub/auth"
+	"github.com/konveyor/tackle2-hub/model"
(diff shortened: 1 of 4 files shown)


jortel (approved):
LGTM

api/adoptionplan.go
```diff
@@ -0,0 +1,245 @@
...
+	var requestedApps []struct {
+		ApplicationID uint `json:"applicationId"`
+	}
+
+	body, err := ioutil.ReadAll(ctx.Request.Body)
+	if err != nil {
+		h.bindFailed(ctx, err)
+	}
+
+	err = json.Unmarshal(body, &requestedApps)
```
jortel, line 55: Why not use Gin bind?

PR #52, opened 2022-03-31
Add repo, binary, identity to app import

mansam:


api/import.go
```diff
@@ -341,24 +341,45 @@ func (h ImportHandler) dependencyFromRow(fileName string, row []string) (app mod
...
 //                            This business service must already exist.
 //
+// Identity: May be left blank, but if provided then an identity of that kind with that name must already exist.
+// Col 6: Kind -- The kind of identity to associate with the application
+// Col 7: Name -- The name of the specific identity to associate with the application.
+//
+// Col 8: Binary -- Binary coordinates.
+//
+// Repository: The following columns are coordinates to a source repository.
+// Col 9: Kind
```
jortel, line 351: Don't need the kind.

api/import.go
```diff
@@ -341,24 +341,45 @@ func (h ImportHandler) dependencyFromRow(fileName string, row []string) (app mod
...
+// Col 10: URL
+// Col 11: Branch
+// Col 12: Tag
```
jortel, line 354: The Tag isn't supported in the UI and should probably be removed from the rest-resource and the model.

api/import.go
```diff
@@ -341,24 +341,45 @@ func (h ImportHandler) dependencyFromRow(fileName string, row []string) (app mod
 // Col 5: Business Service -- The name of the business service this Application should belong to.
 //                            This business service must already exist.
 //
+// Identity: May be left blank, but if provided then an identity of that kind with that name must already exist.
+// Col 6: Kind -- The kind of identity to associate with the application
+// Col 7: Name -- The name of the specific identity to associate with the application.
+//
+// Col 8: Binary -- Binary coordinates.
```
jortel, line 348: Please split the `Binary` field into 4 in the CSV.
Group
Artifact
Version
Packaging

The packaging is optional.  This is stored in Application.Binary joined together:  `<Group>:<Artifact>:<Version>:<Packaging>`.

Please add a validation that when any of: Group, Artifact, Version is specified, they must all 3 be specified.

api/import.go
```diff
@@ -341,24 +341,45 @@ func (h ImportHandler) dependencyFromRow(fileName string, row []string) (app mod
 // Col 5: Business Service -- The name of the business service this Application should belong to.
 //                            This business service must already exist.
 //
+// Identity: May be left blank, but if provided then an identity of that kind with that name must already exist.
+// Col 6: Kind -- The kind of identity to associate with the application
+// Col 7: Name -- The name of the specific identity to associate with the application.
```
jortel, line 346: We don't need support for the identity references.  @rromannissen has said users will be expected to _mass_ associate credentials afterwards in the UI.

model/identity.go
```diff
@@ -10,8 +10,8 @@ import (
 // Identity represents and identity with a set of credentials.
 type Identity struct {
 	Model
-	Kind        string `gorm:"not null"`
-	Name        string `gorm:"not null"`
+	Kind        string `gorm:"uniqueIndex:identity_a;not null"`
+	Name        string `gorm:"uniqueIndex:identity_a;not null"`
```
jortel, line 14: Since we don't need to support unique ref by kind and name, we should remove this.

aufi:
Took over the PR, updated in https://github.com/konveyor/tackle2-hub/pull/74

PR #61, opened 2022-04-14
Wrap errors at the edge to capture stack trace.

jortel:
The design pattern used on MTV and other projects has been to Wrap errors at the edge (where they occur).  The wrap method will capture the stack trace and provide for attaching additional context to the error.  The error is returned. The error is either handled or returned on up the stack by the calling code. Should the error never be handled, it will be logged with the relevant stack trace and attached context.

### DIFF
--- a/addon/client.go
+++ b/addon/client.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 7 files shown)


ibolton336 (approved):
LGTM

PR #64, opened 2022-04-18
Refactor task manager and reaper.

### LINKED ISSUE #40: Limit concurrent task for same addon/application.
Tasks targeted for the same addon and application cannot run concurrently.
Few strategies to consider:
- Enforce uniqueness for Task: addon, application, state=(Created|Ready|Running).
  - Pros: Only one runnable task exists.
  - Cons: difficult to implement.  Don't see how with a constraint which introduces race conditions. Could get around this with _mutex_ in the handler.  Ugh.
- Run tasks with same addon and application serially.
  - Pros: Simple.
  - Cons: tasks could be unintentionally queued up and more difficult for caller to determine which task is theirs.
- New task with same addon and application cancels previous matching task.
  - Pros: Only 1 task in a runnable state for each addon/application.
  - Cons: just need to ask cancel logic to the task manager.

### LINKED ISSUE #63: Reap job for succeeded tasks.
Reap job for succeeded tasks.  This will help keep the job/tasks cleaned up.
Suggested: 12 hours?

jortel:
Added task job reaper.
Add: serialize tasks running against the same addon & application.
Refactor reapers into separate objects in a new `reaper.go`.
Replaced _purge_ terminology with _empty_ (verb).
Renamed `task` package => `tasking`.
Tasks using Pods (instead of Jobs).
Fix issue with using task _name_ for the pod name.

The operator can override here: https://github.com/konveyor/tackle2-operator/issues/47.

closes #40 
closes #63

### DIFF
--- a/tasking/reaper.go
+++ b/tasking/reaper.go
@@ -82,6 +82,7 @@ func (r *TaskReaper) delete(task *model.Task) (deleted bool, err error) {
 	err = r.DB.Delete(task).Error
 	if err == nil {
 		Log.Info("Task deleted.", "id", task.ID)
+		deleted = true
 	} else {
 		err = liberr.Wrap(err)
 	}
@@ -264,6 +265,7 @@ func (r *GroupReaper) delete(g *model.TaskGroup) (deleted bool, err error) {
 	err = r.DB.Delete(g).Error
 	if err == nil {
 		Log.Info("Group deleted.", "id", g.ID)
+		deleted = true
 	} else {
 		err = liberr.Wrap(err)
 	}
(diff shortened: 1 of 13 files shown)


tasking/reaper.go
```diff
@@ -0,0 +1,325 @@
...
+		if err != nil {
+			Log.Trace(err)
+			continue
+		}
+	}
+}
+
+//
+// delete Deletes the task as needed.
+func (r *TaskReaper) delete(task *model.Task) (deleted bool, err error) {
```
mansam, line 78: This appears to always return `deleted=false`

tasking/reaper.go
```diff
@@ -0,0 +1,325 @@
...
+		return
+	}
+	for i := range list {
+		task := &list[i]
+		deleted, err := r.delete(task)
+		if err != nil {
+			Log.Trace(err)
+			continue
+		}
+		if deleted {
```
mansam, line 60: see below regarding `delete()` return value

tasking/reaper.go
jortel: Fixed.

tasking/reaper.go
jortel: Fixed.

ibolton336 (approved):
LGTM

PR #66, opened 2022-04-20
Addon client Resiliant to hub availability.

jortel:
The addon should not fail due to the hub API temporarily not available due to restart or network disruption.
Added the explicit transport configuration.

### DIFF
--- a/addon/client.go
+++ b/addon/client.go
@@ -8,8 +8,10 @@ import (
 	"github.com/konveyor/tackle2-hub/auth"
 	"io"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/url"
+	"time"
 )
 
 //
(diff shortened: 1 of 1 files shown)


addon/client.go
```diff
@@ -179,6 +191,57 @@ func (r *Client) Delete(path string) (err error) {
...
+		return
+	}
+	for i := 0; i < 60; i++ {
+		request, err = rb()
+		if err != nil {
+			return
+		}
+		client := http.Client{Transport: r.transport}
+		response, err = client.Do(request)
+		netErr := &net.OpError{}
```
aufi, line 211: Just a note on the Network Error type, If I understand it correctly, it covers cases with unreachable network, timeout etc., not HTTP errors like 4xx or 5xx - If this is right I assume it was intended.

aufi (approved):

addon/client.go
jortel: Exactly.

PR #69, opened 2022-04-21
Task delete improvements.

jortel:
The problem solved here is:
1. Reduce complexity.  The bucket life-cycle is tied to it's owner but some buckets are shared.
2. Discontinue relying on ORM delete hook to clean up buckets.  Hooks can be temperamental in that they only fire when an operation is done directly with the model.  They do not fire when _Associations_ are managed.
3. The TaskGroup shares it bucket reference with all of its tasks.  The  taskgroup was not being _reaped_ so the bucket was being leaked.

Decouple the bucket life-cycle with its owner (model).  Instead, the bucket owner (Application, TaskGroup and Task) _releases_ their claim to a bucket by no longer referencing it (Bucket="").  Then, have a reaper delete buckets that are no longer referenced.

Another important change is to use `rm -rf` instead of `os.RemoveAll()` which is horribly inefficient on network volumes.


api/application.go
```diff
@@ -135,29 +134,15 @@ func (h ApplicationHandler) Create(ctx *gin.Context) {
...
-			if result.Error != nil {
-				err = result.Error
-				return
-			}
-		}
-		result = db.Select(clause.Associations).Delete(m)
-		if result.Error != nil {
-			err = result.Error
-			return
-		}
```
jortel, line 156: The Application associates can be deleted like other models because we no longer rely on the Task (delete) hook to delete the bucket.

api/application.go
```diff
@@ -184,6 +169,7 @@ func (h ApplicationHandler) Update(ctx *gin.Context) {
 	m.ID = id
 	db := h.DB.Model(m)
 	db = db.Omit(clause.Associations)
+	db = db.Omit("Bucket")
```
jortel, line 172: The read-only permission removed from the model so we need to enforce not updating the bucket ref in the API.

api/task.go
```diff
@@ -219,6 +219,7 @@ func (h TaskHandler) Update(ctx *gin.Context) {
 	db := h.DB.Model(m)
 	db = db.Where("id", id)
 	db = db.Where("state", tasking.Created)
+	db = db.Omit("Bucket")
```
jortel, line 222: The read-only permission removed from the model so we need to prevent updating the bucket ref in the API.

api/taskgroup.go
```diff
@@ -176,6 +176,7 @@ func (h TaskGroupHandler) Update(ctx *gin.Context) {
 			})
 		return
 	}
+	db = db.Omit("Bucket")
```
jortel, line 179: The read-only permission removed from the model so we need to prevent updating the bucket ref in the API.

api/taskgroup.go
```diff
@@ -305,8 +306,7 @@ type TaskGroup struct {
 	Name   string      `json:"name"`
 	Addon  string      `json:"addon"`
 	Data   interface{} `json:"data" swaggertype:"object" binding:"required"`
-	Bucket string      `json:"bucket"`
-	Purged bool        `json:"purged,omitempty"`
```
jortel, line 309: The `purged` field not being used.

hack/cmd/addon/main.go
```diff
@@ -105,7 +105,7 @@ func listDir(d *Data, application *api.Application, paths []string) (err error)
 		err = os.WriteFile(
 			target,
 			b,
-			0644)
+			0666)
```
jortel, line 108: Addons MUST write files in buckets that can be deleted by the hub.

model/bucket.go
```diff
@@ -4,28 +4,15 @@ import (
...
 
 func (m *BucketOwner) BeforeCreate(db *gorm.DB) (err error) {
-	err = m.Create()
-	return
-}
-
-func (m *BucketOwner) BeforeDelete(db *gorm.DB) (err error) {
-	err = m.DeleteBucket()
-	return
-}
```
jortel, line 24: Hook no longer used.
Bucket delete not longer tied to model life-cycle.

model/bucket.go
```diff
@@ -36,25 +23,6 @@ func (m *BucketOwner) Create() (err error) {
...
-//
-// EmptyBucket delete bucket content.
-func (m *BucketOwner) EmptyBucket() (err error) {
-	content, _ := ioutil.ReadDir(m.Bucket)
-	for _, n := range content {
-		p := path.Join(m.Bucket, n.Name())
-		_ = os.RemoveAll(p)
-	}
-	return
-}
```
jortel, line 53: Bucket's no longer emptied but are un-referenced and reaped instead.

settings/hub.go
```diff
@@ -40,7 +40,7 @@ type Hub struct {
 	Task struct {
 		SA      string
 		Retries int
-		Reaper  struct { // hours.
+		Reaper  struct { // minutes.
```
jortel, line 43: Minutes provide finer grained control.  Especially for testing :)

addon/adapter.go
```diff
@@ -25,6 +26,7 @@ var (
 var Addon *Adapter
 
 func init() {
+	unix.Umask(0)
```
jortel, line 29: help support the addon writing files into buckets that can be deleted.

aufi:
Thanks for providing background description via Meet! Notes below:
- The logic added in BucketReaper file that takes care of not-needed Buckets/directories looks great!
- Ack on Bucket not being a DB model, but just a string field on bucket's owner objects specifying filesystem path to the Bucket data (which I somehow expected by default, but makes sense).

Questions (mostly about Release vs. Delete)
- (Task)GroupReaper do a release, when there are existing Tasks, but does not delete itself. Thinking if the release (setting bucket="") might be done in the reaper or sooner (e.g. when Group's Tasks were created).
- Agree there is similar Release concept on a Task with Application - just curious - is a Task without data in Bucket helpful/relevant for its associated Application?

jortel:
> Thanks for providing background description via Meet! Notes below:
> 
> * The logic added in BucketReaper file that takes care of not-needed Buckets/directories looks great!
> * Ack on Bucket not being a DB model, but just a string field on bucket's owner objects specifying filesystem path to the Bucket data (which I somehow expected by default, but makes sense).
> 
> Questions (mostly about Release vs. Delete)
> 
> * (Task)GroupReaper do a release, when there are existing Tasks, but does not delete itself. Thinking if the release (setting bucket="") might be done in the reaper or sooner (e.g. when Group's Tasks were created).

The TaskGroup bucket is released by the reaper immediately the first time the reaper finds the state=Ready (submitted) group.  I considered releasing the bucket as part of the submit but decided that was not a good separation of concerns.

> * Agree there is similar Release concept on a Task with Application - just curious - is a Task without data in Bucket helpful/relevant for its associated Application?

No.

aufi (approved):
Great, thanks for info!

PR #70, opened 2022-04-25
Addon resources section to Addon CRD.

jortel:
Addon resources section to Addon CRD which is used by the task manager when creating Task pods.
For example:
```
spec:
  image: quay.io/jortel/tackle-addon-windup
  mounts:
  - claim: tackle-windup-addon-volume-claim
    name: m2
  resources:
    limits:
      cpu: 500m
      memory: 800Mi
    requests:
      cpu: 100m
      memory: 350Mi
```

This requires a change in the operator.


aufi (approved):
Looks good to me!

PR #75, opened 2022-04-29
Set default RepositoryKind in Application Import

aufi:
Adding RepositoryKind into Application CSV Import and using 'git' as a
default value.


jortel (approved):
LGTM
Thanks!

PR #77, opened 2022-05-01
Add Task.State=Pending to reflect pod pending phase.

jortel:
Add Task.State=Pending to reflect pod pending phase.


aufi (approved):
Looks good to me!

PR #78, opened 2022-05-03
Fix task/secret name for k8s.

jortel:
Fix task/secret name for k8s.  Make them based on Task.ID instead of Task.Name.
Add missing error reporting.
Add missing index on Task.State.

### DIFF
--- a/tasking/manager.go
+++ b/tasking/manager.go
@@ -16,7 +16,6 @@ import (
 	"path"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"strconv"
-	"strings"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


tasking/manager.go
```diff
@@ -128,6 +127,7 @@ func (m *Manager) startReady() {
 				Log.Info("Task postponed.", "id", ready.ID)
 			} else {
 				err := task.Run()
+				Log.Trace(err)
```
aufi, line 130: Just curious if the Log.trace should be called in all cases (regardless to the err), otherwise looks good to me!

aufi (approved):

tasking/manager.go
jortel: Yes, for simplicity.  The Log.Trace() will no-op when err == nil.

PR #79, opened 2022-05-03
Support task priorities.

jortel:
Support task priorities and addon attaching data.

Needed to support https://github.com/konveyor/tackle2-addon/pull/1

### DIFF
--- a/addon/task.go
+++ b/addon/task.go
@@ -154,7 +154,7 @@ func (h *Task) Bucket() (b string) {
 
 //
 // Result report addon result.
-func (h *Task) Result(object interface{}) (err error) {
+func (h *Task) Result(object interface{}) {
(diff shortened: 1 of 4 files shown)


addon/task.go
```diff
@@ -137,32 +137,41 @@ func (h *Task) Completed(n int) {
...
+	b = r.Bucket
+	return
 }
 
 //
-// Bucket returns the bucket path.
-func (h *Task) Bucket() (b string) {
-	r := &api.Task{}
+// Result report addon result.
+func (h *Task) Result(object interface{}) {
```
jortel, line 157: Just the Result is new.
The other diffs are just a result of moving 2 methods.  I like grouping un-exported methods together at the end.

aufi (approved):

PR #80, opened 2022-05-04
Fix task leaks

jortel:
Fix a few leaks:
- The TaskReport not being deleted in the DB when the Reaper deletes tasks.
- The task's secret is not deleted when the task is deleted by Reaper & API.  Added ownerRef.
- The secret or pod can be leaked depending on where the task manager fails to create resources.

### DIFF
--- a/tasking/manager.go
+++ b/tasking/manager.go
@@ -236,8 +236,31 @@ func (r *Task) Run() (err error) {
 		err = liberr.Wrap(err)
 		return
 	}
+	defer func() {
+		if err != nil {
+			_ = r.client.Delete(context.TODO(), &secret)
+		}
+	}()
 	pod := r.pod(&secret)
 	err = r.client.Create(context.TODO(), &pod)
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+	defer func() {
+		if err != nil {
+			_ = r.client.Delete(context.TODO(), &pod)
+		}
+	}()
+	secret.OwnerReferences = append(
+		secret.OwnerReferences,
+		meta.OwnerReference{
+			APIVersion: "v1",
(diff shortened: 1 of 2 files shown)


tasking/reaper.go
```diff
@@ -167,7 +156,11 @@ func (r *TaskReaper) deletePod(m *model.Task) (err error) {
 //
 // delete task.
 func (r *TaskReaper) delete(m *model.Task) {
-	err := r.DB.Delete(m).Error
+	err := r.deletePod(m)
+	if err != nil {
+		Log.Trace(err)
+	}
+	err = r.DB.Select(clause.Associations).Delete(m).Error
```
jortel, line 163: Add the Select(clause.Associations) so the ORM will delete the associated TaskReport.

tasking/reaper.go
```diff
@@ -133,22 +132,12 @@ func (r *TaskReaper) release(m *model.Task) {
 //
 // deletePod Deletes the associated pod as needed.
 func (r *TaskReaper) deletePod(m *model.Task) (err error) {
-	pod := &core.Pod{}
-	err = r.Client.Get(
```
jortel, line 137: No need to Get the pod.

tasking/manager.go
```diff
@@ -236,12 +236,35 @@ func (r *Task) Run() (err error) {
...
+		}
+	}()
+	secret.OwnerReferences = append(
+		secret.OwnerReferences,
+		meta.OwnerReference{
+			APIVersion: "v1",
+			Kind:       "Pod",
+			Name:       pod.Name,
+			UID:        pod.UID,
+		})
```
jortel, line 262: Add the owner ref.

aufi (approved):

PR #83, opened 2022-05-06
Volumes

jortel:
Add _volumes_ as a first-class concept in the hub.
The Addon CR supports defining an RWX volume to be mounted.  Currently the windup addon requires a volume for the m2 cache (local repository).  We have a requirement to:
- Display the capacity of the m2 volume.
- Clean (delete content of) the m2 volume.

To provide this:
- Add `Volume` model.
- Add _volumes/_ endpoint
- Add a _controller_ to the hub to reconcile addons CRs and maintain the volumes in the DB.
- Add _conditions_ to the addon CR status.
- Add _volumes_ service (goroutine) to schedule a task using the _admin_ addon to refresh the volume capacity/used statistics.  While doing this, I separated the existing: reaper and task-manager _services_ into separate packages.  Renamed `tasking` to just `task`.
- Added support for task TTL.

Requires:
- https://github.com/konveyor/tackle2-addon/pull/2
- https://github.com/konveyor/tackle2-operator/pull/56


Makefile
```diff
@@ -9,7 +9,9 @@ PKG = ./addon/... \
       ./k8s/... \
       ./model/... \
       ./settings/... \
-      ./tasking/...
```
jortel, line 12: Renamed for consistency.

addon/error.go
```diff
@@ -47,7 +47,7 @@ type NotFound struct {
 }
 
 func (e NotFound) Error() string {
-	return fmt.Sprintf("GET: path:%s (not-found)", e.Path)
+	return fmt.Sprintf("HTTP path:%s (not-found)", e.Path)
```
jortel, line 50: Noticed the _method_ GET was hard coded and misleading.

addon/volume.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"github.com/konveyor/tackle2-hub/api"
+)
+
+//
+// Volume API.
+type Volume struct {
+	// hub API client.
+	client *Client
+}
```
jortel, line 12: Add to the adapter API for access by addons.

api/addon.go
```diff
@@ -7,7 +7,7 @@ import (
 	crd "github.com/konveyor/tackle2-hub/k8s/api/tackle/v1alpha1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"net/http"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	k8s "sigs.k8s.io/controller-runtime/pkg/client"
```
jortel, line 10: Alias as _k8s_ to avoid name/shadowing collision with `client` variables and parameters.

cmd/main.go
```diff
@@ -116,6 +124,36 @@ func buildScheme() (err error) {
 	return
 }
 
+//
+// addonManager
+func addonManager(db *gorm.DB) (mgr manager.Manager, err error) {
```
jortel, line 129: Builds the addon controller (manager).

controller/addon.go
```diff
@@ -0,0 +1,163 @@
...
+		return
+	}
+
+	// Done
+	return
+}
+
+//
+// changed.
+func (r *Reconciler) changed(addon *api.Addon) (err error) {
```
jortel, line 132: Addon changed.  Delete/Add the volumes in the model

generated/crd/tackle.konveyor.io_addons.yaml
```diff
@@ -86,10 +86,55 @@ spec:
           status:
             description: AddonStatus defines the observed state of Addon
             properties:
+              conditions:
```
jortel, line 89: Add _conditions_ to the status.

model/task.go
```diff
@@ -145,3 +146,14 @@ func (m *TaskGroup) merge(a, b Map) (out Map) {
...
+//
+// TTL time-to-live.
+type TTL struct {
+	Created   int `json:"created,omitempty"`
+	Pending   int `json:"pending,omitempty"`
+	Postponed int `json:"postponed,omitempty"`
+	Running   int `json:"running,omitempty"`
+	Succeeded int `json:"succeeded,omitempty"`
+	Failed    int `json:"failed,omitempty"`
+}
```
jortel, line 159: The TTL is defined for limit to how long a task may exist in each state.

reaper/manager.go
```diff
@@ -1,22 +1,85 @@
-package tasking
+package reaper
```
jortel, line 1: Moved the reaper (service) into a separate package.

controller/addon.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+	}
+
+	return
+}
+
+//
+// deleted.
+func (r *Reconciler) deleted(addon *api.Addon) (err error) {
+	err = r.DB.Delete(&model.Volume{}, "id > 0").Error
```
aufi, line 157: Just curious, this query should delete all (or first of) Volumes with ID > 0?

reaper/manager.go
```diff
@@ -29,70 +92,112 @@ type TaskReaper struct {
...
 //   - Pod is deleted after the defined period.
 //   Failed
-//   - Deleted after the defined period when not associated with
-//     an application. Tasks associated with an application
-//     are deleted when the application is deleted.
+//   - Deleted after TTL > terminated timestamp or
+//     settings.Task.Reaper.Failed.
 //   - Bucket is released after the defined period.
 //   - Pod is deleted after the defined period.
 func (r *TaskReaper) Run() {
```
jortel, line 122: Add support for TTL.
The reaper strategy changed to only delete tasks when a TTL is defined explicitly.  Else, the reaper will only _release_ the task resources.

Note: For analysis tasks, the TTL should never be defined for the Succeeded state.

settings/all.go
```diff
@@ -9,6 +9,7 @@ var Settings TackleSettings
 
 type TackleSettings struct {
 	Hub
+	Metrics
```
jortel, line 12: Needed for the controller manager to avoid TCP port collision.

volume/manager.go
```diff
@@ -0,0 +1,147 @@
+package volume
```
jortel, line 1: New volume manager (service) is responsible for periodically running the task to calculate the capacity/used for volumes.

volume/manager.go
```diff
@@ -0,0 +1,147 @@
...
+//
+// Pause.
+func (m *Manager) pause() {
+	d := Unit * time.Duration(Settings.Frequency.Volume)
+	time.Sleep(d)
+}
+
+//
+// update volumes.
+// The last task is deleted and the task ID reused.
```
jortel, line 67: Attempts to delete/create task with same ID to prevent _eating up_ task IDs.

volume/manager.go
```diff
@@ -0,0 +1,147 @@
...
+	r.Name = r.Variant
+	r.Locator = r.Variant
+	r.Addon = "admin"
+	r.State = task.Ready
+	r.Priority = 1
+	r.TTL = &api.TTL{
+		Pending:   TTL,
+		Succeeded: TTL,
+		Failed:    TTL,
+	}
```
jortel, line 91: Adding the TTL here 2x the frequency to ensure the task is not reaped between runs.

volume/manager.go
```diff
@@ -0,0 +1,147 @@
...
+	if m.taskID == 0 {
+		return
+	}
+	current := &model.Task{}
+	current.ID = m.taskID
+	err = m.DB.First(current).Error
+	if err != nil {
+		return
+	}
+	rt := task.Task{Task: current}
```
aufi, line 128: I'm probably just confused with renaming, but the ```task.Task{}``` is not the ```Task``` struct from addon/task.go (where I don't see the ```Task``` attribute)?

aufi (approved):

controller/addon.go
jortel: Yeah, it's the only way I know of to get the ORM to delete all since IDs begin with 1.

volume/manager.go
jortel: It is the _Runtime_ Task from the _task_ (manager) package.  The _Runtime_ task is an object that represents the Task as it interfaces with k8s.

aufi: Ah, thanks for explanation :+1:

aufi:
LGTM

jortel:
Squashed.

PR #84, opened 2022-05-09
Taskgroup DELETE without transaciton: fixes database locked.

jortel:
Taskgroup DELETE without transaction: fixes `database locked` trace.
The sqlite3 database can encounter "database locked" error when using transactions.  The ORM (gorm) does not protect against this known limitation.

Also, noticed when testing that using the signals package to get the _stop channel_ for the manager makes killing the manager in the debugger annoying.  Not really needed since we don't ever plan to stop the manager.


aufi (approved):
LGTM

PR #86, opened 2022-05-10
Remove url validation

ibolton336:
- Instead of adding a custom validation for git urls, rely on the ui field level validation.


jortel (approved):
LGTM

PR #74, opened 2022-04-28
Update Application CSV import Add repo, binary

aufi:
Taking over Sam's PR https://github.com/konveyor/tackle2-hub/pull/52/ incorporating Jeff's feedback.

This PR extends Application CSV import.

### Testing example

Example CSV file (uses relations to objects created by scripts in hack/add directory), named imp.csv
```
Applications (First line is header)
1,MarketingApplication,My cool app,No comment,Marketing,binarygrp,elfbin,v1,rpm,url,branch,path,Testing,RHEL
```
Upload
```
curl -F "file=@imp.csv" localhost:8080/importsummaries/upload
```

Get imports
```
curl localhost:8080/imports
```
Note: failed binary fields validation sets ```"errorMessage":"Binary-related fields for application MarketingApplication need to be all present or all empty"``` on Import (e.g. if removed "v1" from csv above)

Get applications
```
curl localhost:8080/applications
# ...,{"id":4,"createUser":"","updateUser":"","createTime":"2022-04-28T16:45:20.618059006+02:00","name":"MarketingApplication","description":"My cool app","bucket":"/tmp/bucket/131140af-6833-43e8-a645-674d03abc740","repository":{"kind":"","url":"url","branch":"branch","tag":"","path":"path"},"binary":"binarygrp:elfbin:v1:rpm","facts":null,"review":null,"comments":"No comment","identities":[],"tags":null,"businessService":{"id":1,"name":"Marketing"}}]
```

PR #88, opened 2022-05-12
Add inline Application dependency to CSV import

aufi:
Dependencies were defined in Dependency and DependencyDrection columns in Tackle1.2, the new implementation moved dependencies to own lines (RecordType=2).

Re-adding dependency columns to Application line CSV format (and keeping own lines for dependencies too).

Related to https://github.com/konveyor/tackle2-hub/pull/74


aufi:
Ready for review, works locally.

PR #92, opened 2022-05-12
Refresh ubi base images

fbladilo:
- ubi8 base images cut at 8.4 which has not been updated in 6 months , use latest


jortel (approved):
LGTM

PR #95, opened 2022-05-17
Allow skipping ssl warnings for requests

amastbau:
To reduce the amount of information in the log.


aufi (approved):

aufi:
Looks good, just readme could be updated too.

PR #96, opened 2022-05-18
Ignore import error

amastbau:
Ignore failed uploaded item (due to request failed) and move to the next item


aufi (approved):
Thanks for the PR Amos! There soulc be some follow-up changes, but looks good for now, merging.

PR #100, opened 2022-05-26
TACKLE-633: Add ON DELETE SET NULL to stakeholder FK => job function.

jortel:
Fix http-500 deleting job funcions.
https://issues.redhat.com/browse/TACKLE-633


aufi (approved):
LGTM

PR #103, opened 2022-05-30
Track Username that created or updated models

aufi:
Extending auth provider to get current User's username (```preffered_username``` field from JWT token generated by Keycloak) and setting CreateUser and UpdateUser attributes on models in create&update API calls.

### Development notes
I struggled with gocloak GetUserInfo() method which always failed as part hub's auth middleware on missing or invalid token. Even wrote standalone GetUserInfo-like method in golang directly with net/http https://gist.github.com/aufi/57556376c440301713bde6b04b57ca4a it without gocloak and tested it with curl, however no sucess from the running hub.

Finally, the solution was already part of the existing auth token and its decode method which reads access scopes. So a new method getting the username was added to the auth provider and API Base was updated to call the Username method with token from gin context.

Adding an example on what is currently encoded in the Keycloak token:
image

https://issues.redhat.com/browse/TACKLE-424

### DIFF
--- a/api/base.go
+++ b/api/base.go
@@ -4,18 +4,20 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"reflect"
+	"strconv"
+	"time"
+
 	"github.com/gin-gonic/gin"
 	"github.com/konveyor/tackle2-hub/auth"
 	"github.com/konveyor/tackle2-hub/model"
 	"github.com/mattn/go-sqlite3"
 	"gorm.io/gorm"
-	"io/ioutil"
-	"net/http"
-	"os"
-	"reflect"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"strconv"
-	"time"
 )
 
 //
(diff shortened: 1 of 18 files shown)


api/base.go
```diff
@@ -255,6 +256,19 @@ func (h *BaseHandler) modBody(
 	return
 }
 
+//
+// CurrentUsername gets username from Keycloak auth token
+func (h *BaseHandler) CurrentUsername(ctx *gin.Context) (username string) {
```
jortel, line 261: Just a preference.  `CurrentUser()`.  I generally stay away from _username_ in all cases because it _could_ be confused wth the User's name like "John Doe".

api/base.go
```diff
@@ -255,6 +256,19 @@ func (h *BaseHandler) modBody(
...
+	token := ctx.GetHeader("Authorization")
+	// Consider later add username to ctx and cache for single request?
+	username, err := h.AuthProvider.Username(token)
```
jortel, line 264: Since the middleware has already authenticated the token - how can this fail?

auth/provider.go
```diff
@@ -114,6 +122,29 @@ func (r *Keycloak) newScope(s string) (scope KeycloakScope) {
 	return
 }
 
+//
+// Username resolves token to username using Keycloak service.
+func (r *Keycloak) Username(token string) (username string, err error) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+	defer cancel()
+	decoded, claims, err := r.client.DecodeAccessToken(ctx, token, r.realm)
```
jortel, line 130: Since the middleware has already authenticated the token - how can this timeout? fail?

aufi: It looked to me that once it is a method exposed from the auth package, it should be self-sufficient with error checking, but I agree there is the middeware method being part of the auth pkg which forces the Scopes check - will update lines below keeping just parsing of the "preffered_username" which is not touched by the Scopes method and it depends on Keycloak service config.

api/base.go
aufi: This can fail when the auth token doesn't contain a field with the username. It looks to me to be responsibility of auth pkg to report it as an error, but a responsibility of the api pkg to decide what to do with missing username (setting it to "").

auth/provider.go
jortel: Agreed the method should be self sufficient.  Mainly wanted to understand how it could fail.

PR #104, opened 2022-05-31
Add error for unknown RecordTypes in CSV import

aufi:
QE requested validating RecordTypes during CSV import to keep user up-to-date with amount of records which was accepted/rejected and rejected with unknown type.

The type is processed during initial CSV upload, so adding counter displaying user a error message with information that there is some amount of not accepted records.

https://issues.redhat.com/browse/TACKLE-634


aufi:
Did few experiments, and anbandoning this PR in favor or small UI title change on rejected records.

PR #105, opened 2022-05-31
Fix SEGV in provider and support special task token.

jortel:
Fix SEGV in provider when token is invalid.
Add support for special task token.

### DIFF
--- a/api/base.go
+++ b/api/base.go
@@ -4,8 +4,8 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"github.com/gin-gonic/gin"
+	"github.com/konveyor/controller/pkg/logging"
 	"github.com/konveyor/tackle2-hub/auth"
 	"github.com/konveyor/tackle2-hub/model"
 	"github.com/mattn/go-sqlite3"
@@ -18,6 +18,8 @@ import (
 	"strconv"
 	"time"
 )
+
+var Log = logging.WithName("api")
 
 //
 // BaseHandler base handler.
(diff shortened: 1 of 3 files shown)


api/base.go
```diff
@@ -257,14 +259,12 @@ func (h *BaseHandler) modBody(
...
+// CurrentUser gets username from Keycloak auth token.
 func (h *BaseHandler) CurrentUser(ctx *gin.Context) (user string) {
 	token := ctx.GetHeader("Authorization")
-	// Consider later add username to ctx and cache for single request?
-	user, err := h.AuthProvider.User(token)
+	user, err := auth.CurrentUser(h.AuthProvider, token)
 	if err != nil {
-		fmt.Printf("Failed to get current user, err: %v\n", err)
-		user = ""
+		Log.Error(err, "Failed to get current user.")
```
jortel, line 267: Let's log this instead.

auth/provider.go
```diff
@@ -39,7 +39,8 @@ func (r *NoAuth) Scopes(token string) (scopes []Scope, err error) {
 //
 // User mocks username for NoAuth
 func (r *NoAuth) User(token string) (name string, err error) {
-	return "admin.noauth", nil
+	name = "admin.noauth"
```
jortel, line 42: More consistent when using named return values.

auth/provider.go
```diff
@@ -127,13 +128,13 @@ func (r *Keycloak) newScope(s string) (scope KeycloakScope) {
...
 	defer cancel()
-	// Token validity should be checked before in Scopes method
 	_, claims, err := r.client.DecodeAccessToken(ctx, token, r.realm)
-
-	// Get preferred_username from the token payload as the user
-	user, ok := (*claims)["preferred_username"].(string)
-	if !ok {
-		err = errors.New("cannot parse preferred_username from token")
+	if err != nil {
+		return
```
jortel, line 133: Returns the error. Fixes the SEGV.

aufi (approved):
Thanks for the PR, looks good to me!

PR #109, opened 2022-06-02
Tackle-644: Fixed returned created TagType.

jortel:
https://issues.redhat.com/browse/TACKLE-644

### DIFF
--- a/api/tagtype.go
+++ b/api/tagtype.go
@@ -107,7 +107,7 @@ func (h TagTypeHandler) Create(ctx *gin.Context) {
 	}
 	r.With(m)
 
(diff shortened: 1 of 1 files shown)


api/tagtype.go
```diff
@@ -107,7 +107,7 @@ func (h TagTypeHandler) Create(ctx *gin.Context) {
 	}
 	r.With(m)
 
-	ctx.JSON(http.StatusCreated, m)
+	ctx.JSON(http.StatusCreated, r)
```
jortel, line 110: Just need to return the REST resource instead of the DB model.

ibolton336 (approved):

PR #112, opened 2022-06-07
Add full cleanup to tackle tool

aufi:
Adding clean-all command to support scenario exporting Tackle 1.2 and importing later to some Tackle 2 instance which needs to be cleaned-up first and updating order of import objects which created empty tag types first and then tags with refs to the created tag types.

Also updated README with expected scenarios how the tool should be used.

https://issues.redhat.com/browse/TACKLE-660


jortel (approved):
LGTM

PR #113, opened 2022-06-09
fix err and usage msg

amastbau:


hack/tool/tackle
```diff
@@ -11,7 +11,7 @@ import yaml
 
 parser = argparse.ArgumentParser(description='Konveyor Tackle maintenance tool.')
 parser.add_argument('action', type=str, nargs='*',
-                    help='One or more Tackle commands that should be executed, options: export-tackle1 import clean')
+                    help='One or more Tackle commands that should be executed, options: export-tackle1 import clean clean-all')
```
aufi, line 14: Thanks for adding this :+1:

hack/tool/tackle
```diff
@@ -435,11 +435,11 @@ class Tackle12Import:
                     # Check Application's Assessments first
                     asmnts = apiJSON(self.tackle2Url + "/hub/pathfinder/assessments?applicationId=%d" % importObj['id'], self.tackle2Token, ignoreErrors=True)
                     if len(asmnts) > 0:
-                        print("ERROR: Pathfinder assessment for application ID %d already exists. Clean it before running the import with: tackle clean" % importObj['id'])
+                        print("ERROR: Pathfinder assessment for application ID %d already exists. Clean it before running the import with: tackle clean or tackle clean-all" % importObj['id'])
```
aufi, line 438: In the preImportCheck method, there should be only the "clean" command as a solution of duplicates.

The clean-all might delete more objects than necessary and also might skip / leave existing some of assessments (thise without currently existing application). I tried describe the difference in https://github.com/konveyor/tackle2-hub/tree/main/hack/tool#if-the-import-failed

I'm not saying that the clean-all will never work for your use case, likely it will, but I wanted make the clean operation as not-invasive for user as possible, so it deletes only objects with are in collision with the imported data.

hack/tool/tackle
```diff
@@ -435,11 +435,11 @@ class Tackle12Import:
...
                         exit(1)
                 for destObj in destCollection:
                     if importObj['id'] == destObj['id']:
-                        print("ERROR: Resource %s/%d \"%s\" already exists in Tackle2 destination as \"%s\". Clean it before running the import with: tackle clean" % (t, importObj['id'], importObj['name'], destObj['name']))
+                        print("ERROR: Resource %s/%d \"%s\" already exists in Tackle2 destination as \"%s\". Clean it before running the import with: tackle clean or tackle clean-all" % (t, importObj['id'], importObj['name'], destObj['name']))
```
aufi, line 442: The same, please keep just the clean command.

aufi (changes requested):

aufi (approved):
Looks good to me.

PR #118, opened 2022-06-23
Update CSV import always count from zero

aufi:
Updating Application CSV import dependency syntax to support only the full import format (without shorthands for dependencies), that removes counting back from the last dependency column.

Addressing technical debt issue: https://issues.redhat.com/browse/TACKLE-610

### DIFF
--- a/api/import.go
+++ b/api/import.go
@@ -314,8 +314,8 @@ func (h ImportHandler) DownloadCSV(ctx *gin.Context) {
 // Col 1: Record Type 1 -- This will always contain a "2" for a dependency
 // Col 2: Application Name -- The name of the application that has the dependency relationship.
 //                            This application must exist.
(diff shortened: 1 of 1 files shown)


api/import.go
```diff
@@ -324,22 +324,13 @@ func (h ImportHandler) DownloadCSV(ctx *gin.Context) {
 // Examples:
 //
 // 2,MyApplication,,,,OtherApplication,NORTHBOUND,,,,,,,,
-// 2,MyApplication,OtherApplication,SOUTHBOUND
```
aufi, line 327: This "shorter" way of defining dependencies is not possible anymore, empty columns needs to be used. That should now work as expected (from discussion with Ramon) and fits to CSV template provided in 2.0 release.

aufi:
Tested locally with the template CSV provided by Tackle2-UI.

PR #119, opened 2022-06-24
Create missing BusinessService in CSV import

aufi:
When there is an unknown BusinessService in Application CSV import, it will be created automatically.


aufi:
There will be an upcoming PR covering this together with createEntities option and TagTypes&Tags creation.

aufi:
Change introduced in this PR is part of https://github.com/konveyor/tackle2-hub/pull/123

PR #120, opened 2022-06-27
Fix tool export Stakeholders

aufi:
There was a bug with import/export tool exporting Stakeholders and their connection to StakeholderGroups, fixed.


gildub (approved):
LGTM

PR #121, opened 2022-06-27
Update tool README with requirements

aufi:
Adding Python and PyYAML requirement into tackle tool README, updates making it working on Python3, but <= 3.9, few minor usage updates and skip Tackle2 token request when using --skip-destination-check option.

Now should work on Python 3.6 (RHEL8) out-of-the box and Python 3.9 (RHEL9) with manually installed PyYAML.

### DIFF
--- a/hack/tool/tackle
+++ b/hack/tool/tackle
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import argparse
 import copy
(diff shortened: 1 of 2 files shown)


hack/tool/tackle
```diff
@@ -74,17 +74,16 @@ def getKeycloakToken(host, username, password, client_id='tackle-ui', realm='tac
 
 def apiJSON(url, token, data=None, method='GET', ignoreErrors=False):
     debugPrint("Querying: %s" % url)
-    match method:
```
aufi, line 77: The match-case works in Python >=3.10, rewritten to if-elif-else to ensure compatibility with older Python versions.

hack/tool/tackle
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
```
aufi, line 1: In case, there is not symlink python->python3, using python3 explicitly. (RHEL)

scoonRH (commented):
Looks good.  Thanks Marek!

PR #122, opened 2022-06-30
Update tool pair Tags case insensitive

aufi:
There was a problem with Tags - "Postgresql" in Tackle 1.2, "PostgreSQL" in Tackle 2.0. The pairing code which should migrate just Tags which are not already present in Tackle 2.0 consideted these to be a different tags. However sqlite prevented creation of the "Postgresql" tag when there was "PostgreSQL" alredy present in the same TagType.

Adding ```lower()``` method to Tag names to tackle tool to match tags case insensitive, so it considers "Postgresql" same as "PostgreSQL".

Export/import of whole Tackle 1.2 with -s option was not affected by this problem since all tags from 1.2 were exported and all seed tags from 2.0 are deleted before import.

Fixes: https://issues.redhat.com/browse/TACKLE-683

### DIFF
--- a/hack/tool/tackle
+++ b/hack/tool/tackle
@@ -154,13 +154,13 @@ class Tackle12Import:
         for tt2 in collection:
             tt  = Tackle2Object(tt2)
             tt.name = tt2['name']
-            self.destData['tagtypes'][tt.name] = tt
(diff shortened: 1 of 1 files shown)


jortel (approved):
LGTM

PR #124, opened 2022-07-19
Add --no-auth to Tackle tool

aufi:
Adding --no-auth (-n shortcut) flag to Tackle CLI tool. If used, Keycloak token creation is skipped and empty token is used in Tackle API calls.

Needed for RHPDS setup.


jortel (approved):
LGTM

PR #125, opened 2022-07-19
Cascade application delete to pathfinder.

jortel:
Cascade application delete to pathfinder. Need to complete: https://issues.redhat.com/browse/TACKLE-673.
The UI will no longer need to explicitly delete of assessments when an application is deleted.

Requires: new pathfinder _bulkDelete_ endpoint.


mansam (approved):
LGTM

PR #127, opened 2022-07-25
Introduce data migration mechanism

mansam:
This adds a mechanism for automatically performing
versioned data migrations, as well as schema migrations
that GORM can't do automatically.

This also adds a migration for 2.1.0 which makes the
Application -> BusinessService relationship optional
by changing the constraint to ON DELETE SET NULL.

Updated the version of GORM to pick up a fix for a bug
in the sqlite driver that would prevent a table from
being recreated correctly.


database/pkg.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+//
+// Perform any necessary DB migrations.
+func migrate(db *gorm.DB) (err error) {
+	var migrations = []func(*gorm.DB) error{
+		v210,
+	}
+
+	db.Exec("PRAGMA foreign_keys = OFF")
```
mansam, line 129: GORM compensates for sqlite not natively supporting changing constraints by creating a new table, copying data, and dropping the old one. We have to shut off the foreign key pragma first, otherwise this will result in data being lost.

aufi (approved):
LGTM, but Jeff should approve too.

mansam:
Jeff and I had a long discussion about this approach, and we settled on a different design that I'm currently prototyping.

PR #128, opened 2022-07-28
Fix autocreation of tags

mansam:
* Skip the tagtype and tag when the tag name is empty.
* Return an error if the tagtype is empty but tag name is present.
* Fix name normalization to only strip `-_\s`

### DIFF
--- a/importer/manager.go
+++ b/importer/manager.go
@@ -185,6 +185,17 @@ func (m *Manager) createApplication(imp *model.Import) (ok bool) {
 		// Prepare normalized names for importTag
 		normImpTagName := normalizedName(impTag.Name)
 		normImpTagType := normalizedName(impTag.TagType)
+
+		// skip if tag name normalizes to an empty string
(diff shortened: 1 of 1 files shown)


jortel (approved):
LGTM

importer/manager.go
```diff
@@ -185,6 +185,17 @@ func (m *Manager) createApplication(imp *model.Import) (ok bool) {
...
 		normImpTagName := normalizedName(impTag.Name)
 		normImpTagType := normalizedName(impTag.TagType)
+
+		// skip if tag name normalizes to an empty string
+		if normImpTagName == "" {
+			continue
+		}
+		// fail if the tag name is ok but the tag type normalizes to an empty string
+		if normImpTagType == "" {
+			imp.ErrorMessage = fmt.Sprintf("Tag '%s' has missing or invalid TagType.", impTag.Name)
```
jortel, line 195: The "_or invalid_" part may be misleading.

PR #129, opened 2022-08-08
Allow empty App-BusinessService link in CSV import

aufi:
Application CSV import have BusinessService field which could be empty and that should not result to creation a new BusinessService record with empty name, but should make an Application without linked BusinessService.

Part of code making BusinessService association on Application with pointer comes from @mansam's closed PR https://github.com/konveyor/tackle2-hub/pull/127

Fixes https://issues.redhat.com/browse/TACKLE-719

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -308,7 +308,7 @@ type Application struct {
 	Comments        string      `json:"comments"`
 	Identities      []Ref       `json:"identities"`
 	Tags            []Ref       `json:"tags"`
-	BusinessService Ref         `json:"businessService"`
+	BusinessService *Ref        `json:"businessService"`
 }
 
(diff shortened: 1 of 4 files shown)


aufi:
The update of BusinessService model needs to be agreed by Jeff and Sam before merging.

jortel (changes requested):
LGTM
Blocking merge pending @mansam PoC re: migration/upgrade mechanism.  If we go with the upgrade mechanism, this can be merged to main.  Else it would need to be merged to a special 2.1.0-for-MTA6 branch.

aufi:
Replaced with https://github.com/konveyor/tackle2-hub/pull/140

PR #130, opened 2022-08-10
Introduce version migration mechanism

mansam:
We cannot rely on GORM to completely automigrate a sqlite database, and doing so poses a problem both for seeding the
database on a fresh install and for updating an existing deployment of Tackle. To avoid these problems, @jortel and I
discussed always starting a fresh Tackle install with only the Setting table, and then applying migrations to the schema
one by one (rather than having GORM autocreate all tables, which it does currently).

In order to avoid having to write DDL for each schema version, we keep a history of each database model in versioned
packages. At each step of the migration process, we can then use GORM's AutoMigrate to apply the appropriate set of
models for the version, manually apply any changes AutoMigrate can't do on its own, and then seed required models into
the tables. This minimizes the amount of work that AutoMigrate needs to do, and gives us more precise control over
schema migration. Once a migration is applied to the database, a Setting record is updated with the name of the applied version.
The migration is performed as a single transaction so that any failure will cause the changes to be rolled back to the last successful version.


database/pkg.go
```diff
@@ -0,0 +1,72 @@
...
+	"gorm.io/gorm"
+	"gorm.io/gorm/schema"
+)
+
+var log = logging.WithName("db")
+
+var Settings = &settings.Settings
+
+const (
+	HubDBVersion     = ".hub.db.version"
```
jortel, line 20: Thinking this may not be seeded here.
Also, that the version WRT the migration would not be DB only.

database/pkg.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+	err = db.AutoMigrate(model.Setting{})
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+	result := db.FirstOrCreate(&model.Setting{}, model.Setting{Key: HubDBVersion})
```
jortel, line 53: Thinking it would be more decoupled to not be seeded here.

migration/pkg.go
```diff
@@ -0,0 +1,111 @@
...
+func All() []Migration {
+	return []Migration{
+		{Version: "v2.0.0", Apply: v200},
+		{Version: "v2.1.0", Apply: v210},
+	}
+}
+
+//
+// Migrate the hub by applying all necessary Migrations.
+func Migrate(migrations []Migration) (err error) {
```
jortel, line 53: Would think the impl would be in migrate.go (or whatever) and not in the pkg.go.

migration/pkg.go
```diff
@@ -0,0 +1,111 @@
...
+	liberr "github.com/konveyor/controller/pkg/error"
+	"github.com/konveyor/controller/pkg/logging"
+	"github.com/konveyor/tackle2-hub/database"
+	"github.com/konveyor/tackle2-hub/model"
+	"gorm.io/gorm"
+)
+
+var log = logging.WithName("migration")
+
+const HubDBVersion = ".hub.db.version"
```
jortel, line 15: Recommend this be (something like):
Version = "hub.migration.version" or just ".migration.version"

migration/pkg.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+//
+// Migrate the hub by applying all necessary Migrations.
+func Migrate(migrations []Migration) (err error) {
+	for _, m := range migrations {
+		var db *gorm.DB
+		db, err = database.Open(false)
+		if err != nil {
+			log.Error(err, "Failed to open database for migration.")
```
jortel, line 58: Let's consider wrap and return. 
```
liberr.Wrap(err, "version", version)
```
Also, missing a return here.

migration/v210.go
```diff
@@ -0,0 +1,29 @@
+package migration
+
+import (
+	liberr "github.com/konveyor/controller/pkg/error"
+	"github.com/konveyor/tackle2-hub/model"
+	"gorm.io/gorm"
+)
+
+func v210(db *gorm.DB) (err error) {
+	err = db.AutoMigrate(model.All()...)
```
jortel, line 10: This needs to be associated to a `v210` package not latest, right?

model/v200/pkg.go
```diff
@@ -0,0 +1,101 @@
...
+		TaskGroup{},
+		TaskReport{},
+		Volume{},
+		Proxy{},
+	}
+}
+
+//
+// Seed the database with models.
+func Seed(db *gorm.DB) (err error) {
```
jortel, line 43: Perhaps put this in seed.go instead of pkg.go?

migration/pkg.go
```diff
@@ -0,0 +1,34 @@
...
+// Version represents the value of the .migration.version setting.
+type Version struct {
+	Version int `json:"version"`
+}
+
+//
+// Migration encapsulates the functionality necessary to perform a migration.
+type Migration interface {
+	Apply(*gorm.DB) error
+	DisplayName() string
```
jortel, line 24: One pedantic request:  Even though we discussed this as a "display name", I generally think it's better for fields/methods not be named for how they are used and instead just the what they are.
As such would recommend: `Name()`.

migration/migrate.go
```diff
@@ -0,0 +1,94 @@
...
+
+//
+// Migrate the hub by applying all necessary Migrations.
+func Migrate(migrations []Migration) (err error) {
+	var db *gorm.DB
+	db, err = database.Open(false)
+	if err != nil {
+		return
+	}
+
```
jortel, line 19: I wonder if we don't want something like:
```
defer func() {
    if err != nil {
        _ = database.Close(db)
    }
}
```
here to prevent leaking db connections on error?

The `var db *gorm.DB` on line 47 would need to be removed so the loop is just re-assigning the outer `db`.

migration/migrate.go
```diff
@@ -0,0 +1,101 @@
...
+		m := migrations[i]
+
+		db, err = database.Open(false)
+		if err != nil {
+			err = liberr.Wrap(err, "version", m.Name())
+			return
+		}
+
+		f := func(db *gorm.DB) (err error) {
+			log.V(3).Info("Running migration.", "version", m.Name())
```
jortel, line 61: I think log.Info() (not level=3) would be better.

jortel (approved):
LGTM

PR #136, opened 2022-08-19
Make Application->BusinessService rel optional

mansam:

PR #140, opened 2022-08-24
Update CSV import to optional BusinessService

aufi:
Updating CSV application import code to count with optional relation Application-BusinessService (instead of mandatory).

Related to: https://github.com/konveyor/tackle2-hub/pull/136


mansam (approved):
LGTM

PR #126, opened 2022-07-21
Add Tackle2 export to CLI tool

aufi:
Adding Tackle 2 API objects export to tackle CLI tool and few minor updates on the tool.

The Tackle 2 export scenario added to README:
```
### Export and Import data between Tackle 2 instances

This feature focuses on application inventory, but exports and imports most of Tackle 2 objects except TaskGroups&Tasks and Keycloak identitty provider data.

Export/imported entities:
- Applications, Dependencies, Assessments, Reviews, Tag Types, Tags, Job Functions, Stakeholder Groups, Stakeholders, Business Services, Proxies, Identities (sensitive data are encrypted with ```encryption_passphase``` config file field)
- Application's Buckets data (e.g. app binaries or windup reports)

PR #146, opened 2022-09-07
[WIP] Add bucket data GET&PUT to API

aufi:
Bucket is a directory containing filesystem tree structure, there was no API method allowing download/upload bucket data (that would be helpful e.g. in Tackle 2 export&import).

Adding Bucket operations GET (download tar.gz archive of bucket content) and PUT (upload tar.gz archive with content which should replace given bucket data).

Related to export/import Tackle 2 including bucket data PR: https://github.com/konveyor/tackle2-hub/pull/126

### DIFF
--- a/api/bucket.go
+++ b/api/bucket.go
@@ -1,17 +1,211 @@
 package api
 
 import (
-	"github.com/gin-gonic/gin"
-	"github.com/konveyor/tackle2-hub/model"
+	"archive/tar"
+	"bufio"
+	"bytes"
+	"compress/gzip"
+	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"os"
 	pathlib "path"
+	"path/filepath"
+	"regexp"
+
+	"github.com/gin-gonic/gin"
+	"github.com/konveyor/tackle2-hub/model"
+)
+
+//
+// Routes
+const (
+	BucketsRoot = "/buckets"
+	BucketRoot  = BucketsRoot + "/:" + ID
 )
 
 //
 // BucketHandler provides bucket management.
 type BucketHandler struct {
+	BaseHandler
+}
+
+//
+// AddRoutes adds routes.
+func (h BucketHandler) AddRoutes(e *gin.Engine) {
+	routeGroup := e.Group("/")
+	// TODO: add auth middleware when buckets scope added to the token
+	// routeGroup.Use(auth.AuthorizationRequired(h.AuthProvider, "buckets"))
+	routeGroup.GET(BucketRoot, h.Get)
+	routeGroup.PUT(BucketRoot+"/:"+ID, h.Put)
+}
+
+// Get godoc
+// @summary Upload bucket content archive by ID.
+// @description Upload a bucket content .tar.gz archive by ID as PUT (replace all previous bucket content).
+// @tags put
+// @produce octet-stream
+// @success 200 {binary}
+// @router /buckets/{id} [put]
+// @param id path string true "Bucket ID"
+func (h BucketHandler) Put(ctx *gin.Context) {
+	invalidSymbols := regexp.MustCompile("[^a-zA-Z0-9-_]")
+	bucketID := invalidSymbols.ReplaceAllString(ctx.Param("ID"), "")
+	bucketPath := "/tmp/bucket" + bucketID
+
+	// Prepare to uncompress the uploaded data
+	file, err := ctx.FormFile("file")
+	if err != nil {
+		ctx.Status(http.StatusBadRequest)
+		return
+	}
+	fileReader, err := file.Open()
+	if err != nil {
+		ctx.Status(http.StatusBadRequest)
+		return
+	}
+	defer fileReader.Close()
+
+	ungzReader, err := gzip.NewReader(fileReader)
+	if err != nil {
+		ctx.Status(http.StatusBadRequest)
+		return
+	}
+	defer ungzReader.Close()
+
+	// Clean the destination bucket directory (or do an atomic way - extract to some temp directory and move when extreact finished successfully)
+	bucketContent, err := ioutil.ReadDir(bucketPath)
+	if err != nil {
+		log.Error(err, "read bucket dir")
+		ctx.Status(http.StatusInternalServerError)
+		return
+	}
+	for _, bucketEntry := range bucketContent {
+		err = os.RemoveAll(bucketPath + "/" + bucketEntry.Name())
+		if err != nil {
+			log.Info("Cleaning bucket dir, cannot delete")
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


api/bucket.go
```diff
@@ -1,19 +1,205 @@
...
+}
+
+func (h *BucketHandler) serveBucketUpload(ctx *gin.Context, owner *model.BucketOwner) {
+	if ctx.Request.Method == "PUT" {
+		h.uploadDirArchive(ctx, path.Join(owner.Bucket, ctx.Param(Wildcard)))
+	} else {
+		h.upload(ctx, owner)
+	}
+}
+
```
aufi, line 43: The uploadDirArchive method takes a gin ctx with an uploaded ```file``` pointing to the tar.gz archive and a string specifying a destination directory where the archive should be extracted to. The directory needs to exist before and extration replaces all previous content.

api/bucket.go
```diff
@@ -1,19 +1,205 @@
...
+			}
+			file.Close()
+		default:
+			// types that are not files or dirs are skipped
+		}
+	}
+
+	ctx.Status(http.StatusAccepted)
+}
+
```
aufi, line 120: The getDirArchive method takes a gin ctx as a destination where it sends the streamed archive data and a string specifying a directory that should be archived.

api/bucket.go
```diff
@@ -1,19 +1,205 @@
...
+func (h *BucketHandler) serveBucketGet(ctx *gin.Context, owner *model.BucketOwner) {
+	if ctx.Request.Header.Get(Accept) == TarGzMimetype {
+		h.getDirArchive(ctx, path.Join(owner.Bucket, ctx.Param(Wildcard)))
+	} else {
+		h.content(ctx, owner)
+	}
+}
+
+func (h *BucketHandler) serveBucketUpload(ctx *gin.Context, owner *model.BucketOwner) {
+	if ctx.Request.Method == "PUT" {
```
aufi, line 37: There is was a problem with mimetype, the file upload set the HTTP request to be ```multipart/form-data``` which wraps the uploaded tar.gz file. The API use POST&PUT for the same method (https://github.com/konveyor/tackle2-hub/blob/main/api/application.go#L39-L40), so it looked acceptable to use the PUT method in condition wheter a single file should be uploaded (as until now) or a file with tar.gz archive should be extracted to given path.

This approach can be discussed, but it looks that checking Content-Type of the PUT request for application/x-gzip doesn't seem to work.

api/bucket.go
```diff
@@ -1,19 +1,205 @@
...
+	"github.com/konveyor/tackle2-hub/model"
 )
 
 //
 // BucketHandler provides bucket management.
 type BucketHandler struct {
 }
 
+// TODO: Validate Wildcard param to not allow access dirs outside of the bucket
+func (h *BucketHandler) serveBucketGet(ctx *gin.Context, owner *model.BucketOwner) {
```
aufi, line 28: These two methods are called from the owner APIs. Their responsibility is to decide if a single file or an archive should be uploaded/downloaded and potentialy also validation the ```Wildcard``` parameter to not escape the bucket path.

aufi:
It is a time to submit this PR (even symlinks&permissions were taken out since being WIP) for review to get feedback on the re-usable methods for uploading/downloading directory archives as tar.gz files.

api/application.go
```diff
@@ -243,15 +243,15 @@ func (h ApplicationHandler) Update(ctx *gin.Context) {
 // @router /applications/{id}/tasks/{id}/content/{wildcard} [get]
 // @param id path string true "Task ID"
 func (h ApplicationHandler) BucketGet(ctx *gin.Context) {
-	taskID := h.pk(ctx)
+	id := h.pk(ctx)
```
jortel, line 246: :+1:

aufi:
Sample curl for for application bucket.

Get archive: ```curl -vOJ -H "Accept: application/x-gzip" localhost:8080/applications/1/bucket/```

Upload archive and extract it to the bucket directory: ```curl -v -X PUT -F "file=@bucket.tar.gz" localhost:8080/applications/1/bucket/```

api/bucket.go
```diff
@@ -1,19 +1,218 @@
...
+		}
+
+		switch hdr.Typeflag {
+		case tar.TypeDir:
+			if err := os.Mkdir(path.Join(dir, hdr.Name), 0777); err != nil {
+				return
+			}
+		case tar.TypeReg:
+			var file *os.File
+			if file, err = os.Create(path.Join(dir, hdr.Name)); err != nil {
```
aufi, line 111: Permissions are 0666 before applying umask by default from https://pkg.go.dev/os#Create

api/bucket.go
```diff
@@ -1,19 +1,209 @@
...
 )
 
 //
 // BucketHandler provides bucket management.
 type BucketHandler struct {
 }
 
+func (h *BucketHandler) serveBucketGet(ctx *gin.Context, owner *model.BucketOwner) {
+	if ctx.Request.Header.Get(Directory) == DirectoryArchive {
+		h.getDirArchive(ctx, path.Join(owner.Bucket, ctx.Param(Wildcard)))
```
aufi, line 28: Was thinking about the Wildcard validation to not allow "applications/1/bucket/../.." GET or PUT, but this directory traversal is handled by gin router which "resolves" ../.. as part of URL parsing which results to "applications/".

api/bucket.go
```diff
@@ -1,19 +1,209 @@
...
+			return
+		}
+	}()
+
+	// Clean and prepare destination directory
+	err = os.RemoveAll(dir)
+	if err != nil {
+		return
+	}
+	if err = os.Mkdir(dir, 0777); err != nil {
```
jortel, line 79: I think we'll need os.MakedirAll() here to ensure intermediate directories are created.

api/bucket.go
```diff
@@ -1,19 +1,209 @@
...
+			})
+			return
+		}
+	}()
+
+	if err != nil {
+		return
+	}
+
+	if err := tarWriter.Close(); err != nil {
```
jortel, line 189: Should this be in a defer instead?

aufi: The Close() method writes footer to the tar archive (it "closes" the archive content, not the stream), it needs to be executed before ```io.Copy(gzWriter, fromTar)``` on line 202.

jortel (approved):
Few comments.
Request a regression test with UI with one of the following use cases:
- Run analysis with uploaded binary.
- Run analysis with custom rules file.

LGTM

aufi:
Successfully tested analysis on an uploaded war binary (with custom tackle2-hub mage built from main&this PR ```quay.io/maufart/tackle2-hub:bucketmain```).

PR #151, opened 2022-09-30
Add /schema endpoint.

jortel:
Add /schema endpoint.
We needed to add an unauthenticated endpoint to support Hub (pod) readiness. May as well make it return something useful.

**Needed for Tackle 2.1.2.**


mansam (approved):
LGTM

PR #152, opened 2022-10-03
Avoid cyclic dependencies

aufi:
Adding GORM BeforeCreate Hook for Dependencies to avoid creating cyclic Application Dependencies.

Example: having dependencies apps 1->2, 2->3, then 3->1 should not be allowed, but 1->3 is still OK.

Related to https://issues.redhat.com/browse/TACKLE-814

Signed-off-by: Marek Aufart <maufart@redhat.com>

### DIFF
--- a/api/base.go
+++ b/api/base.go
@@ -30,16 +30,13 @@ type BaseHandler struct {
 	Client client.Client
 	// Auth provider
 	AuthProvider auth.Provider
-	// Synchronization helpers (with mutex)
-	SyncHelper *model.SyncHelper
 }
 
 // With database and k8s client.
-func (h *BaseHandler) With(db *gorm.DB, client client.Client, provider auth.Provider, syncHelper *model.SyncHelper) {
+func (h *BaseHandler) With(db *gorm.DB, client client.Client, provider auth.Provider) {
 	h.DB = db.Debug()
 	h.Client = client
 	h.AuthProvider = provider
-	h.SyncHelper = syncHelper
 }
 
 //
--- a/api/base.go
+++ b/api/base.go
@@ -30,13 +30,16 @@ type BaseHandler struct {
 	Client client.Client
 	// Auth provider
 	AuthProvider auth.Provider
+	// Synchronization helpers (with mutex)
+	SyncHelper *model.SyncHelper
 }
 
 // With database and k8s client.
-func (h *BaseHandler) With(db *gorm.DB, client client.Client, provider auth.Provider) {
+func (h *BaseHandler) With(db *gorm.DB, client client.Client, provider auth.Provider, syncHelper *model.SyncHelper) {
 	h.DB = db.Debug()
 	h.Client = client
 	h.AuthProvider = provider
+	h.SyncHelper = syncHelper
 }
 
 //
@@ -101,6 +104,11 @@ func (h *BaseHandler) createFailed(ctx *gin.Context, err error) {
 			sqlite3.ErrConstraintPrimaryKey:
 			status = http.StatusConflict
 		}
+	}
+
+	depCyclicError := &model.DependencyCyclicError{}
+	if errors.As(err, depCyclicError) {
+		status = http.StatusConflict
 	}
 
 	ctx.JSON(
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -97,6 +97,7 @@ func addonManager(db *gorm.DB, adminChanged chan int) (mgr manager.Manager, err 
 func main() {
 	log.Info("Started", "settings", Settings)
 	var err error
+	syncHelper := model.SyncHelper{}
 	defer func() {
 		if err != nil {
 			log.Trace(err)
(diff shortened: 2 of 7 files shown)


migration/v2/model/application.go
```diff
@@ -105,3 +108,24 @@ type ImportTag struct {
 	ImportID uint `gorm:"index"`
 	Import   *Import
 }
+
+//
+// Validation Hook to avoid cyclic dependencies.
+func (dep *Dependency) BeforeCreate(db *gorm.DB) (err error) {
```
aufi, line 114: The GORM Hook on the model level should cover both API&CSV import use cases.

migration/v2/model/application.go
```diff
@@ -105,3 +108,24 @@ type ImportTag struct {
...
+	var nextDeps []*Dependency
+	var nextAppsIDs []uint
+	nextAppsIDs = append(nextAppsIDs, dep.FromID)
+	for len(nextAppsIDs) != 0 {
+		db.Where("ToID IN ?", nextAppsIDs).Find(&nextDeps)
```
aufi, line 119: Amount of SQL queries grows from 1 by "level of depth" of dependency tree (the reason for using ```IN``` query). This looked acceptable for me, but I'm happy to incorporate more efficient solution if recommended one.

migration/v2/model/application.go
jortel: What about update _"Manage Dependencies"_ use cases?

jortel:
I wonder how we deal with race conditions?
2+ api clients adding dependencies at the same time.
Would it be a bad idea to add a MUTEX here?

jortel:
Let's also make sure we _bubble up_ an error that will will result in the API returning a 400 instead of a 500.

aufi:
Thanks for feedback! Added mutex for synchronizing Dependency create in CSV import and Dependency create in API and mustom error type for cyclic dependency error returning 409 error.

cmd/main.go
```diff
@@ -97,6 +97,7 @@ func addonManager(db *gorm.DB, adminChanged chan int) (mgr manager.Manager, err
 func main() {
 	log.Info("Started", "settings", Settings)
 	var err error
+	syncHelper := model.SyncHelper{}
```
aufi, line 100: In order to keep the same mutex in Gin.based API and backend Import manager, it was needed add the mutex to "top level" of the application, which maybe results in slightly bigger change than expected.

jortel (changes requested):
Perhaps, I am missing something but - I don't see why we'd need to introduce a new interface and concept of SyncHelper It seems unnecessarily complicated.  I _think_ this kind of thing should be rare and dealt with individually. A _synchronized_ method is a common design pattern.
Why not just create a package local Mutex like:
```
//
// depMutex ensures Dependency.Create() is not executed concurrently.
var depMutex sync.Mutex
...
...
///
/// Create a dependency synchronized with a mutex.
func (r *Dependency) Create(db *gorm.DB) (err error) {
    depMutex.Lock()
    defer depMutex.Unlock()
    err = db.Create(r)
    return
}

func (r *Dependency) BeforeCreate(db *gorm.DB) (err error) {
    ...
```
The APIs can be changed to:
```
m.Create(h.DB)
```
instead of:
```
h.DB.Create(m)
```

The Create() and BeforeCreate() (hook) could be combined but probably better separate in case the Create() is not used, the check will still be performed.

Also, I would really prefer we never name things _helper_ or _util_.  IMHO, all things should be understood well enough to have a descriptive name.

aufi:
The initial idea was to share the Mutex between API and processImport package. Sharing ```depMutex``` in via cmd/main.go file looked awkward to me, so I wrapped it into the syncHelper for possibly similar use cases.

Thanks for recommending create the Dependency model Create method and using it for syncing purpose!

> The APIs can be changed to:
> 
> ```
> m.Create(h.DB)
> ```
> 
> instead of:
> 
> ```
> h.DB.Create(m)
> ```
> 
> The Create() and BeforeCreate() (hook) could be combined but probably better separate in case the Create() is not used, the check will still be performed.
> 

Changed both API and processImport Dependency creation to the reversed create command (model-DB).

> Also, I would really prefer we never name things _helper_ or _util_. IMHO, all things should be understood well enough to have a descriptive name.

OK, makes sense :+1:

migration/v2/model/application.go
```diff
@@ -105,3 +112,45 @@ type ImportTag struct {
 	ImportID uint `gorm:"index"`
 	Import   *Import
 }
+
+//
+// Create a dependency synchronized using a mutex.
+func (r *Dependency) Create(db *gorm.DB) (err error) {
+	depMutex.Lock()
+	defer depMutex.Unlock()
+	result := db.Create(r)
```
jortel, line 121: err = db.Create(r).Error

api/base.go
```diff
@@ -103,6 +103,11 @@ func (h *BaseHandler) createFailed(ctx *gin.Context, err error) {
 		}
 	}
 
+	depCyclicError := &model.DependencyCyclicError{}
+	if errors.As(err, depCyclicError) {
+		status = http.StatusConflict
```
jortel, line 108: :+1: Conflict instead of Bad-Request.

migration/v2/model/application.go
aufi: Updated :+1:

jortel (approved):
LGTM

jortel:
Thanks for putting up with all my requests.

PR #154, opened 2022-10-04
TACKLE-654: Better support DB concurrency.

jortel:
As recommended by sqlite3 driver [documentation](https://pkg.go.dev/github.com/mattn/go-sqlite3#section-readme), we need to:
- Set the `_journal=WAL` in the DSN (better performance).
- Limit the open connections to (1).

The included unit test reproduced the _database locked_ error.  This patch eliminates the error.

https://issues.redhat.com/browse/TACKLE-654

### DIFF
--- a/database/db_test.go
+++ b/database/db_test.go
@@ -1,0 +1,39 @@
+package database
+
+import (
+	"encoding/json"
+	"fmt"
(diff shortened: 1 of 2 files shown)


mansam (approved):
LGTM

PR #162, opened 2022-10-10
Configure Keycloak users and roles during startup

mansam:


settings/auth.go
```diff
@@ -12,17 +12,27 @@ const (
...
+	KeycloakRealm        = "konveyor"
+	KeycloakClientID     = "konveyor"
+	KeycloakClientSecret = "konveyor"
+	AddonToken           = "konveyor"
+	KeycloakAdminUser    = "admin"
+	KeycloakAdminPass    = "admin"
+	KeycloakAdminRealm   = "master"
+	RolePath             = "/tmp/roles.yaml"
+	UserPath             = "/tmp/users.yaml"
 )
```
jortel, line 36: In other settings, we don't have const for defaults but define them in the load() instead.
The thinking here is that we don't outside code to accedentally (or intentionally) use the default instead of the _Setting_.
I realize you did not introduce the _const_ in the PR but would recommend it would be a good opportunity to address it.

auth/provider.go
```diff
@@ -56,26 +61,280 @@ func (r *NoAuthScope) Allow(_ string, _ string) (ok bool) {
...
+//
+// ensureUsers ensures that the hub users exist and have the necessary roles.
+func (r *Keycloak) ensureUsers(realm *Realm) (err error) {
+	users, err := LoadUsers(Settings.Auth.UserPath)
+	if err != nil {
+		return
+	}
+
+	for _, user := range users {
+		if _, ok := realm.Users[user.Name]; !ok {
```
jortel, line 161: This is just a personal nit: but I find the `ok` convention in Go rather odd and un-descriptive and feel that something more descriptive like `found` would be better.

auth/role.go
```diff
@@ -0,0 +1,72 @@
+package auth
```
jortel, line 1: docstrings?

auth/provider.go
```diff
@@ -56,26 +61,280 @@ func (r *NoAuthScope) Allow(_ string, _ string) (ok bool) {
...
+			_true := true
```
jortel, line 162: Perhaps `enabled` would be better than `_true`

auth/provider.go
```diff
@@ -56,26 +61,280 @@ func (r *NoAuthScope) Allow(_ string, _ string) (ok bool) {
...
+		}
+		var realmRoles []gocloak.Role
+		for _, role := range user.Roles {
+			realmRole, ok := realm.Roles[role]
+			if !ok {
+				continue
+			}
+			realmRoles = append(realmRoles, realmRole)
+		}
+		err = r.client.AddRealmRoleToUser(context.Background(), r.token.AccessToken, r.realm, *realm.Users[user.Name].ID, realmRoles)
```
jortel, line 189: Will this fail if the user already has one or more of the roles?

auth/provider.go
```diff
@@ -56,26 +61,280 @@ func (r *NoAuthScope) Allow(_ string, _ string) (ok bool) {
...
+			}
+		}
+	}
+
+	// create scope mappings and add default scopes to client
+	idOfClient, err := r.idOfClient(r.id)
+	if err != nil {
+		return
+	}
+	for sid, roles := range scopesToRoles {
```
jortel, line 247: Will `CreateClientScopesScopeMappingsRealmRoles()` and `AddDefaultScopeToClient()` fail of one or more of the mappings already exist?

auth/provider.go
mansam: No, this does not appear to fail.

auth/provider.go
mansam: No, this does not appear to fail.

jortel (approved):
LGTM

auth/provider.go
```diff
@@ -56,26 +61,333 @@ func (r *NoAuthScope) Allow(_ string, _ string) (ok bool) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*3)
+	defer cancel()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+			r.token, err = r.client.LoginAdmin(ctx, r.admin, r.pass, r.adminRealm)
+			if err != nil {
+				time.Sleep(time.Second)
```
jortel, line 385: Do we want to log something here?

PR #168, opened 2022-10-13
Create Keycloak realm and client.

mansam:
Signed-off-by: Sam Lucidi <slucidi@redhat.com>


auth/provider.go
```diff
@@ -151,6 +163,63 @@ func (r *Keycloak) loadRealm() (realm *Realm, err error) {
 	return
 }
 
+//
+// ensureRealm ensures that the hub realm exists.
+func (r *Keycloak) ensureRealm() (err error) {
+	_, err = r.client.GetRealm(context.Background(), r.token.AccessToken, r.realm)
+	if err != nil {
+		if strings.HasPrefix(err.Error(), "404") {
```
jortel, line 171: Error string parsing.  Yikes.

auth/provider.go
```diff
@@ -151,6 +163,63 @@ func (r *Keycloak) loadRealm() (realm *Realm, err error) {
...
+			displayName := fmt.Sprintf("%s realm", r.realm)
```
jortel, line 172: The realm suffix in the display name seems redundant but guess it's what was used by the original tackle team?

jortel (approved):
LGTM

PR #185, opened 2022-11-17
Addon binding client improvements.

jortel:
Addon binding client improvements mainly focused on supporting addon writers running the addon locally.

- Add `Accept` header to avoid 502 by nginx reverse proxy when Accept not set.  Helpful for addon development.
- Append the endpoint path to the Hub URL path instead of overwriting it. Need for Helpful for addon development which needs to set HUB_BASE_URL to the route/ingress:  Eg: http://192.168.49.2/hub.  The /hub needs to be preserved.
- Set the token in the Authentication header in send(). Duplicated.
- Fix logic in send when and error is returned by `client.Do()`.  Fixes SEGV.

### DIFF
--- a/addon/client.go
+++ b/addon/client.go
@@ -12,7 +12,13 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	pathlib "path"
 	"time"
+)
+
+const (
+	Accept  = "Accept"
+	AppJson = "application/json"
 )
 
 //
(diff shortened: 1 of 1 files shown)


mansam (approved):
LGTM

addon/client.go
```diff
@@ -183,7 +189,7 @@ func (r *Client) Delete(path string) (err error) {
 			Method: http.MethodDelete,
 			URL:    r.join(path),
 		}
-		request.Header.Set(auth.Header, r.token)
+		request.Header.Set("Accept", "")
```
aufi, line 192: Assuming intention not to mix the const with a string value.

jortel: good catch.  Nope just missed this spot.

PR #187, opened 2022-12-01
Fix hack addon tag management.

jortel:
Fix hack addon tag management.
- Ensures type exists.
- Ensures tag exists.

Updates the README to running locally.

Add `run-addon` target to the makefile.


jmontleon (approved):
ACK

jmontleon (approved):

PR #188, opened 2022-12-08
Smart addon image pull policy based on image fqin.

jortel:
To better support development/testing: Adjust addon image pull policy based on fqin.
The pod image is format is: `[registry/]repository:tag`
When the _repository_ includes a '/' then it is remote and the pull policy is _Always_.  Else, _Never_.


jortel:
Can't determine if the image references a host.

jortel:
Can't determine if the image references a host.

PR #191, opened 2022-12-19
Update swagger schema files

aufi:
Based on @mguetta1 input, updated swagger annotations to specify optional body on Task/TaskGroup Submit action. Swagger files re-generation added also Schema endpoint into the spec.

Signed-off-by: Marek Aufart <maufart@redhat.com>


api/schema.go
```diff
@@ -46,3 +42,8 @@ func (h *SchemaHandler) Get(ctx *gin.Context) {
 
 	ctx.JSON(http.StatusOK, schema)
 }
+
+type Schema struct {
```
aufi, line 46: In order to pass the docs generation (swagger spec files), moved the Schema type to be a first-class citizen in api package.

ibolton336 (approved):

api/schema.go
aufi: Confirmed with Jeff to be OK.

Issue #193, opened 2023-01-01
bug: HTML <li> tag not closed in the example addon code


HarikrishnanBalagopal:
https://github.com/konveyor/tackle2-hub/blob/a63c8eb5d1b2bbe6bd016842115b0d5c8b04bd74/hack/cmd/addon/main.go#L152

jortel:
Fix merged.

PR #194, opened 2023-01-01
fix: the closing tag was missing from the li element in html of the example addon code

HarikrishnanBalagopal:
Signed-off-by: Harikrishnan Balagopal <harikrishmenon@gmail.com>

### DIFF
--- a/hack/cmd/addon/main.go
+++ b/hack/cmd/addon/main.go
@@ -149,7 +149,7 @@ func buildIndex(output string) (err error) {
 	for _, name := range list {
 		body = append(
 			body,
-			"<li><a href=\""+name.Name()+"\">"+name.Name()+"</a>")
+			"<li><a href=\""+name.Name()+"\">"+name.Name()+"</a></li>")
 	}
 
 	body = append(body, "</ul>")


HarikrishnanBalagopal:
Fixes: #193

aufi:
Thanks for the fix @HarikrishnanBalagopal! May I ask you for not changing formatting (removal of empty comments and import order change)?

HarikrishnanBalagopal:
> May I ask you for not changing formatting (removal of empty comments and import order change)?

It's Golang fmt command https://pkg.go.dev/fmt that runs automatically when you save the file.

aufi (approved):
Thanks! Looks good to me.

aufi:
> > May I ask you for not changing formatting (removal of empty comments and import order change)?
> 
> It was done by the Golang `fmt` command https://pkg.go.dev/fmt that runs automatically when you save the file in VSCode. I have made the changes you requested.

@HarikrishnanBalagopal  It seems most devs use GoLand which formates it in that way. Actually I use VScode too, it is possible adjust Format Tool in go extension settings. I'm not sure it is the best solution, but somehow works for me.
image

PR #218, opened 2023-01-31
Refactor migration v3 to import unchanged models

mansam:
* Remove v1 migration and models
* Organize DB models as one model per file
* Imports models that have not changed from the v2 migration package to the v3 package
* Add check for earliest supported version and fail migration if the DB version is too early.
* Add `Development` setting, and automigrate models on start if set to true even if the db version is up to date.

Signed-off-by: Sam Lucidi <slucidi@redhat.com>


migration/pkg.go
```diff
@@ -2,18 +2,24 @@ package migration
...
+var Settings = &settings.Settings
 
 //
 // VersionKey is the setting containing the migration version.
 const VersionKey = ".migration.version"
 
+//
+// SupportedFrom is the index of the
+// earliest version that we can migrate from.
+const SupportedFrom = 1
```
jortel, line 21: If we still need the `nil` in the ALL slice, why do we need this ^.  Wondering if the migrator could just skip the nil migrations.  Less to keep in sync.

migration/migrate.go
```diff
@@ -45,39 +46,56 @@ func Migrate(migrations []Migration) (err error) {
...
 
 		err = database.Close(db)
 		if err != nil {
-			err = liberr.Wrap(err, "version", m.Name())
+			err = liberr.Wrap(err, "version", ver)
+			return
+		}
+	}
+
+	if Settings.Hub.Development {
```
jortel, line 95: :+1:

cmd/main.go
```diff
@@ -37,7 +37,7 @@ func init() {
 //
 // Setup the DB and models.
 func Setup() (db *gorm.DB, err error) {
-	err = migration.Migrate(migration.All())
+	err = migration.Migrate(migration.All(), migration.SupportedFrom)
```
jortel, line 40: I'm curious about why this is passed in since it's managed/owned by the migration package?  The setting can be manipulated for testing.

jortel: Also (don't mean to be picky but) I wonder if somethiing like `minVersion` would be _better_?

jortel (approved):
LGTM

PR #221, opened 2023-02-02
:sparkles: Add file support.

jortel:
Support uploading and storing files independently of buckets.

Resource:
```
//
// File REST resource.
type File struct {
    Resource
    Name       string     `json:"name"`
    Path       string     `json:"path"`
    Expiration *time.Time `json:"expiration,omitempty"`
}
```

API:
- POST|PUT /files/_name_  - Create|update file. Returns `File`
- GET /files - List files. Returns `[]File`
- GET /files/:id - Get file content.  Accept=json returns `File`, else returns file stream.
- DELETE /files/:id - Delete a file.

Binding:
- File.Get(id uint, destination) - Get (downlaod) a file written to `destination`.
- File.Put(source) - Put (uploads) a file. Returns a `File`.
- File.Delete(id uint) - Delete a file.

Files with Zero reference count are reaped after 12 hours. **Q: Is this too long?**

---

Use case.  The upcoming `RuleSet` needs to have an associated image file.  The anticipated workflow is:

1. Update an Image.
2. Create the RuleSet with RuleSet.Image = File ID returned by the POST on /files.

Should the flow be interrupted between 1 & 2, the orphaned file will be reaped.

Example:
```
type RuleSet struct {
    Model
    ImageID uint `ref:"file"`
    Image   *File
}
```

### DIFF
--- a/api/file.go
+++ b/api/file.go
@@ -3,7 +3,7 @@ package api
 import (
 	"github.com/gin-gonic/gin"
 	"github.com/konveyor/tackle2-hub/auth"
-	"github.com/konveyor/tackle2-hub/migration/v3/model"
+	"github.com/konveyor/tackle2-hub/model"
 	"io"
 	"mime"
 	"net/http"
--- a/api/file.go
+++ b/api/file.go
@@ -5,8 +5,10 @@ import (
 	"github.com/konveyor/tackle2-hub/auth"
 	"github.com/konveyor/tackle2-hub/migration/v3/model"
 	"io"
+	"mime"
 	"net/http"
 	"os"
+	pathlib "path"
 	"time"
 )
 
@@ -133,6 +135,10 @@ func (h FileHandler) Get(ctx *gin.Context) {
 	if result.Error != nil {
 		h.reportError(ctx, result.Error)
 		return
+	}
+	header := ctx.Writer.Header()
+	header[ContentType] = []string{
+		mime.TypeByExtension(pathlib.Ext(m.Name)),
 	}
 	ctx.File(m.Path)
 }
--- a/api/file.go
+++ b/api/file.go
@@ -43,7 +43,7 @@ func (h FileHandler) AddRoutes(e *gin.Engine) {
 // @description List all files.
 // @tags get
 // @produce json
-// @success 200 {object} []api.FileField
+// @success 200 {object} []api.File
 // @router /files [get]
 func (h FileHandler) List(ctx *gin.Context) {
 	var list []model.File
@@ -68,9 +68,9 @@ func (h FileHandler) List(ctx *gin.Context) {
 // @tags create
 // @accept json
 // @produce json
-// @success 201 {object} api.FileField
+// @success 201 {object} api.File
 // @router /files [post]
-// @param name path string true "FileField name"
+// @param name path string true "File name"
 func (h FileHandler) Create(ctx *gin.Context) {
 	var err error
 	input, err := ctx.FormFile(FileField)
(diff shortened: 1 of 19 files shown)


api/file.go
```diff
@@ -0,0 +1,193 @@
+package api
+
+import (
+	"github.com/gin-gonic/gin"
+	"github.com/konveyor/tackle2-hub/auth"
+	"github.com/konveyor/tackle2-hub/migration/v3/model"
```
mansam, line 6: This should import from the model package

api/file.go
```diff
@@ -0,0 +1,193 @@
...
+// FileHandler handles file routes.
+type FileHandler struct {
+	BaseHandler
+}
+
+//
+// AddRoutes adds routes.
+func (h FileHandler) AddRoutes(e *gin.Engine) {
+	routeGroup := e.Group("/")
+	routeGroup.Use(auth.Required("files"))
```
mansam, line 32: This will require that new permissions are added to the role yaml.

api/file.go
```diff
@@ -0,0 +1,193 @@
...
+	routeGroup.GET(FileRoot, h.Get)
+	routeGroup.DELETE(FileRoot, h.Delete)
+}
+
+// List godoc
+// @summary List all files.
+// @description List all files.
+// @tags get
+// @produce json
+// @success 200 {object} []api.FileField
```
mansam, line 46: The doc strings all refer to `api.FileField` where they mean to refer to the `api.File` resource type.

api/file.go
```diff
@@ -0,0 +1,193 @@
...
+	m := &model.File{}
+	m.Name = ctx.Param(ID)
+	m.CreateUser = h.BaseHandler.CurrentUser(ctx)
+	result := h.DB.Create(&m)
+	if result.Error != nil {
+		h.reportError(ctx, result.Error)
+		return
+	}
+	defer func() {
+		if err != nil {
```
mansam, line 90: Should we log these file errors?

api/file.go
```diff
@@ -0,0 +1,193 @@
...
+	m := &model.File{}
+	id := h.pk(ctx)
+	result := h.DB.First(m, id)
+	if result.Error != nil {
+		h.reportError(ctx, result.Error)
+		return
+	}
+	header := ctx.Writer.Header()
+	header[ContentType] = []string{
+		mime.TypeByExtension(pathlib.Ext(m.Name)),
```
mansam, line 141: :+1:

migration/v3/model/file.go
```diff
@@ -0,0 +1,33 @@
...
+	Name       string
+	Path       string `gorm:"<-:create;uniqueIndex"`
+	Expiration *time.Time
+}
+
+func (m *File) BeforeCreate(db *gorm.DB) (err error) {
+	uid := uuid.New()
+	m.Path = path.Join(
+		Settings.Hub.Bucket.Path,
+		".file",
```
mansam, line 23: What's the reason for the dot in the `/.file/` component of the path?

reaper/file.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+//
+// busy determines if anything references the file.
+func (r *FileReaper) busy(file *model.File) (busy bool, err error) {
+	nRef := int64(0)
+	var n int64
+	ref := RefCounter{DB: r.DB}
+	for _, m := range []interface{}{
+		// ADD MODELS HERE
```
mansam, line 68: :+1:

mansam (changes requested):
Looks good, just a few small comments.

migration/v3/model/file.go
jortel: It's stored in the bucket root and . prefixed files are ignored by the bucket reaper.  This can be changed when I update the buckets to be more like files.

mansam (approved):
LGTM, and running the hack addon was successful.

PR #229, opened 2023-02-10
:sparkles: Push multi-arch builds to quay

jmontleon:
Signed-off-by: Jason Montleon <jmontleo@redhat.com>

As a first step to automating releases I'd like to merge to each of the tackle repos a GitHub action to build multi-arch images for the main branch and tags. I've opened up a commit here for review and once we agree I will submit a PR to each of the tackle repos.

For a preview of what this looks like on quay see: https://quay.io/repository/jmontleon/tackle2-hub?tab=tags

Once those are up and they're approved we can turn off the quay auto builds, ensure the robot env vars are in the org, and the account is privileged to push to each repo.

Future improvements from here will include using manifest SHAs in the tag version CSVs and automating the release, ensuring we push a no arch bundle in our quay org, an updated index in our quay org, and auto generating a PR against community operators.

Things we should decide on:
Branch names: do we want to use branches like `release-x.y`, `tackle-x.y`, something else.  
Tag names: Are we good with `vx.y.z` or do we want something else like `x.y.z`, or other.

### DIFF
--- a/.github/workflows/march-image-build-push.yml
+++ b/.github/workflows/march-image-build-push.yml
@@ -1,0 +1,29 @@
+name: 'Build and Push Multi-Arch Image'
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - 'main'
+      - 'release-*'
+    tags:
+      - 'v*'
+
+jobs:
+  push-quay:
+    name: Build and Push Manifest
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+    steps:
+    - name: Checkout Push to Registry action
+      uses: konveyor/release-tools/build-push-quay@main
+      with:
+        architectures: "amd64, arm64, ppc64le, s390x"
+        containerfile: "./Dockerfile"
+        image_name: "tackle2-hub"
+        image_namespace: "konveyor"
+        image_registry: "quay.io"
+        quay_robot: ${{ secrets.QUAY_PUBLISH_ROBOT }}
+        quay_token: ${{ secrets.QUAY_PUBLISH_TOKEN }}
+        ref: ${{ github.ref }}


.github/workflows/march-image-build-push.yml
```diff
@@ -0,0 +1,78 @@
...
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - 'main'
+      - 'release-*'
+    tags:
+      - 'v*'
+  schedule:
+    - cron: '0 0 * * *'
```
shawn-hurley, line 12: Is this saying to build a new multi-arch image every day at midnight?

.github/workflows/march-image-build-push.yml
```diff
@@ -0,0 +1,78 @@
...
+    steps:
+    - name: Checkout Push to Registry action
+      uses: actions/checkout@main
+
+    - name: Install latest podman
+      if: matrix.install_latest
+      run: |
+        bash .github/install_latest_podman.sh
+
+# This package is not new enough for ppc64le builds
```
shawn-hurley, line 34: Should we remove this if we can use the docker action?

shawn-hurley (commented):
+1 on making this standard!

I vote for release- not tackle- I am also voting for vX.Y.Z for the tags 

Can we make sure to add it to the java-analyzer-bundle as well?

.github/workflows/march-image-build-push.yml
jmontleon: Yes, removed! It works fine with the docker action alternative.

.github/workflows/march-image-build-push.yml
jmontleon: Yes. I don't think there is a need for this; I may have carelessly copied it from something else I was working on (konveyor/builder) as a starting point.

jmontleon:
> +1 on making this standard!
> 
> I vote for release- not tackle- I am also voting for vX.Y.Z for the tags
> 
> Can we make sure to add it to the java-analyzer-bundle as well?

I will make sure we add it to java-analyzer-bundle as well!

djzager (commented):
I like this a lot. The only thing I wonder is if we should pull out the `push-quay` to konveyor/release-tools as a "composite" action or something so that, in the future, changes to this logic don't need to be made individually to each repo that relies on it.

jmontleon:
> I like this a lot. The only thing I wonder is if we should pull out the `push-quay` to konveyor/release-tools as a "composite" action or something so that, in the future, changes to this logic don't need to be made individually to each repo that relies on it.

I think we cannot use secrets in a composite action from reading, or is my understanding wrong? My concern on seeing that is passing the credentials to push using `redhat-actions/push-to-registry`.

If my understanding is correct perhaps a reusable workflow would be a valid approach that achieves similar. My understanding is secrets can be used with those.

jmontleon:
Maybe I'm misunderstanding and they just need to be passed as parameters. I'll keep reading...

.github/workflows/march-image-build-push.yml
```diff
@@ -0,0 +1,34 @@
...
+      fail-fast: false
+    steps:
+    - name: Checkout Push to Registry action
+      uses: konveyor/release-tools/build-push-quay@main
+      with:
+        architectures: "amd64, arm64, ppc64le, s390x"
+        containerfile: "./Dockerfile"
+        image_name: "tackle2-hub"
+        image_namespace: "konveyor"
+        image_registry: "quay.io"
```
djzager, line 31: Do you want to use the `env.IMAGE_*` here? or just drop the `env`?

djzager (changes requested):
One nit I can see. Otherwise lgtm.

djzager (approved):

PR #233, opened 2023-02-14
:sparkles: Bucket GET on directory using accept header.

jortel:
Signed-off-by: Jeff Ortel <jortel@redhat.com>

Update the bucket APIs to return tarball for directories based on `Accept` header instead of `X-Directory` header. This will be simpler for browsers and file download widgets.

When the path is a directory in the bucket a tarball will be returned unless the client explicitly accepts `text/html`.

This should not require any changes in API clients (Like the CLI).

### DIFF
--- a/addon/client.go
+++ b/addon/client.go
@@ -229,7 +229,6 @@ func (r *Client) BucketGet(source, destination string) (err error) {
 			Method: http.MethodGet,
 			URL:    r.join(source),
 		}
-		request.Header.Set(api.Directory, api.DirectoryArchive)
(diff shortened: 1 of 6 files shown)


aufi (commented):
Watching the change in order to update export/import CLI as needed.

aufi (approved):
LGTM

Seems to work fine for headers like
```accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7```

PR #235, opened 2023-02-16
:bug: Fix number as json root; custom json field no longer used.

### LINKED ISSUE #227: Fix json field number values.
**Problem**:
The GORM JSON datatype has a bug where it does not support the json document root being a _number_. As a result, the value cannot be a number.  The ORM will store the value but fails to unmarshall it when data is queried.

**Proposal**:
Discontinue using the GORM JSON type.  Replace the alias `JSON = datatypes.JSON` with  `JSON = []byte`. To ensure long term options, we'll want to be sure to set the `gorm:"field:json"` tag on all JSON fields.  This will ensure we can roll our own custom datatype or return to using the one provided by GORM (after it has been fixed) in the future.
This will be backwards compatible.  No migration required. Let's fix this in v2 & v3.

We'll at least file an issue with GORM and perhaps a patch.

jortel:
fixes #227

### DIFF
--- a/migration/v2/model/application.go
+++ b/migration/v2/model/application.go
@@ -6,10 +6,10 @@ type Application struct {
 	Name              string `gorm:"index;unique;not null"`
 	Description       string
 	Review            *Review `gorm:"constraint:OnDelete:CASCADE"`
-	Repository        JSON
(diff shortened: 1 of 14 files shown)


mansam (approved):
LGTM

PR #237, opened 2023-02-20
:ghost: Fix file reaping of rulebundle images and ruleset files.

jortel:
Needed to use IsExported() instead of CanSet().
Other minor:
- passing model as pointer for efficiency.
- add/update logging.
- comments.


aufi (approved):
LGTM

PR #245, opened 2023-02-27
:sparkles: Add `source` to ApplicationTags jointable

mansam:


addon/application.go
```diff
@@ -103,23 +104,62 @@ func (h *Application) Tags(id uint) (tg AppTags) {
...
+func (h *AppTags) Source(name string) {
+	h.source = &name
+}
+
+//
+// Replace the associated tags for the source with a new set.
+// Returns an error if the source is not set.
+func (h *AppTags) Replace(ids []uint) (err error) {
+	if h.source == nil {
+		err = errors.New("`source` must be set")
```
jortel, line 121: liberr.New() to capture the context.

api/application.go
```diff
@@ -677,9 +801,14 @@ func (r *Application) With(m *model.Application) {
...
 			ref)
 	}
-	for _, tag := range m.Tags {
-		ref := Ref{}
-		ref.With(tag.ID, tag.Name)
+}
+
+//
+// WithTags updates the resource with the associated tags.
+func (r *Application) WithTags(tags []model.ApplicationTag) {
```
jortel, line 808: Curious why this is separate instead of:
```
With(m *model.Application, tags []model.ApplicationTag)
```

jortel (approved):
LGTM

api/application.go
mansam: Couldn't decide what felt cleaner. Happy to make it one method like that.

PR #246, opened 2023-02-27
:sparkles: Add bucket directory download content filtering.

### LINKED ISSUE #243: Add support for filtering of GET bucket directory.
Add support for filtering of GET bucket directory.
For example: /application/1/bucket/windup/report?filter=*.csv

jortel:
Add bucket directory download content filtering. Support optional `?filter=<glob>` query parameter.

Example:  /applications/1/buckets/windup/report?*.csv the tarball will only contain .csv files.

Removed the optimization of returning NoContent when the tarball is empty.  I think this complicates what the client needs to do for no real gain. 

closes #243


mansam (approved):
LGTM

Issue #228, opened 2023-02-10
Tag associations not being cleaned up on application delete.


jortel:
It's been reported that Tag associations not being cleaned up on application delete.
Steps:
1. Create application foo.
2. Associate tags.
3. Delete foo.
4. Create application foo.
5. Associate tags (see conflict).

PR #248, opened 2023-02-28
:bug: Fix cascading deletes on tags and applications

mansam:
Fixes cascading delete behavior on the ApplicationTags association table. Also adds cascading delete for Facts so that the presence of a Fact doesn't prevent the Application from being deleted.

Fixes https://github.com/konveyor/tackle2-hub/issues/228


go.mod
```diff
@@ -19,9 +19,9 @@ require (
 	gorm.io/datatypes v1.0.5
 	gorm.io/driver/mysql v1.3.5 // indirect
 	gorm.io/driver/postgres v1.2.3 // indirect
-	gorm.io/driver/sqlite v1.3.6
+	gorm.io/driver/sqlite v1.4.4
 	gorm.io/driver/sqlserver v1.2.1 // indirect
-	gorm.io/gorm v1.23.8
+	gorm.io/gorm v1.24.5
```
mansam, line 24: The GORM update allegedly contains a bugfix for cascading delete behavior for many2many relationships.

PR #252, opened 2023-03-02
:ghost: fix null issues relating to tag associations

mansam:
*Fixes a null pointer while generating the list of tag refs for the resource to be returned from an Application create.
*Fixes a null constraint violation caused by GORM not omitting associations properly while updating.


jortel (approved):
LGTM

PR #255, opened 2023-03-03
:bug: Fix issues with updating models that have associations

mansam:
Only the last call to `Omit` is honored when building a query, so change all multiple uses to a single use with multiple parameters.

Also, exclude more fields from `BaseHandler.fields`, which is used when determining which fields to updaate:
* Pointers to struct
* Arrays and slices
* Pointers to arrays and slices


jortel (approved):
LGTM

PR #258, opened 2023-03-06
:bug: Fix broken import of tagged applications

mansam:
The import manager was preloading the `TagCategory` relationship for Tags, but the association field is named `Category`. This was causing Categories to not be preloaded on the Tag models, which makes imports fail. (either by failing to find the category and giving up, or by trying to recreate the already-existing category depending on whether `createEntities` is enabled.). 

This change fixes import of tagged application by preloading the right association, and ensuring that ApplicationTag associations are created properly. I also took the opportunity to simplify (to my eyes) some of the variable naming.

### DIFF
--- a/importer/manager.go
+++ b/importer/manager.go
@@ -194,13 +194,14 @@ func (m *Manager) createApplication(imp *model.Import) (ok bool) {
 	}
 
 	// Process import Tags & TagCategories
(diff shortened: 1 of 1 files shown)


aufi (approved):
LGTM

jortel (approved):
LGTM

PR #261, opened 2023-03-08
:bug: Don't omit BucketID when creating tasks

mansam:
Don't use `TaskHandler.omitted()` when creating a Task through the API. Doing so was causing the `BucketID` that was being set by the `BeforeCreate` hook to not be persisted on the new Task row in the database. (The in-memory ID for the orphaned bucket was being returned in the response to the `POST`, but subsequent `GET`s would not not include it and the bucket would be reaped.)

`omitted()` is to ensure that fields that are not allowed to be updated through the API (and are thus not propagated from the resource to the model) are not zeroed out when the task is updated.


jortel (approved):
LGTM

Issue #262, opened 2023-03-09
Add testing suite with coverage of the REST API


dymurray:
## Summary ##
Borrowing the goals from https://github.com/konveyor/tackle2-hub/discussions/241, we want to implement regression testing for the hub API by providing a suite of tests that will guarantee the hub conforms to the published API. This suite should be extensible so that new features can include test coverage for any API changes.

The details of this suite should be published in enhancements as follow-ups to https://github.com/konveyor/enhancements/pull/98.

## Proposed Design ##

The API tests should be configurable to run disconnected (outside of a cluster) or connected. When running disconnected, tests requiring a cluster will be omitted:
- Run tasks.
- List/Get addon

### Packages ###
- **test/** - Provides tests.
- **test/api/** - Provides API tests.
- **/test/api/_resource_/**  - Each _resource_ (package) is a top level resource that roughly equals an API endpiont.
- **/test/api/fixtures/**  - Fixtures.  Mainly imported from hub/api and hub/addon to begin with.
- **test/binding/** - Provides addon binding tests.
- **test/_TBD_** - Future.

Example packages:
**/test/api/application** - contains one (or more .go files) Ex: `api_test.go`

### Fixtures ###

The test/api package will provide fixtures (mainly imported from other packages in the hub).
- **Resources** (objects) from the `api` package.
  - api.Application
  - api.Tag
  - api.Task
  - api.JobFunction
  - ...
- _Client_ - REST client (imported from `addon` _binding_).
- _Param_ - Support query parameters (imported from `addon` _binding_).
- _Params_ - Support URL parameters (imported from `addon` _binding_).

Example `test/api/fixtures/client.go`
```

var (
    // Client is the global configured client.
    Client *addon.Client
)

func init() {
    Client = addon.NewClient(settings.Addon.URL, settings.Addon.Token)
}
```

Example usage: `test/api/application/api_test.go`:
```
import (
    api github/konveyor/tackle2-hub/test/api/fixtures
) 

var (
    client = api.Client
}

func TestApplication() {
    //
    // Create
    created := &api.Application{
        Name: "Test",
        Description: "My test."
        ....
    }
    err := client.Post(api.ApplicationsRoot, created)
    //
    // Get
	found := &api.Application{}
	path := api.Params{api.ID: created.ID}.inject(api.ApplicationRoot)
	err = client.Get(path, found)
    //
    // Update
    created.Description = "Updated"
    err = client.Put(path, created)
    //
    // Get
[...]

PR #268, opened 2023-03-17
:seedling: Initial Hub API Tests

aufi:
First part of Hub REST API test suite.

Steps
- the simplest CRUD for all resource types
- simple seeds tests
- connecting resources with refs (update tests above)
- clarify non-trivial fixtures definition and tear up&down
- basic integration scenarios (addons&analysis)

Related to https://github.com/konveyor/tackle2-hub/issues/262


test/api/application/create_test.go
```diff
@@ -0,0 +1,70 @@
+package application
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/konveyor/tackle2-hub/api"
+)
+
+func TestApplicationCreate(t *testing.T) {
```
jortel, line 10: Looks like every test function has an TestApplication prefix that is redundant in the `application` package.  Everything in the package is implied to be scoped to application.
Recommend:
For example: just `TestCreate()` instead of: `TestApplicationCreate()`.  (less stuttering please).

test/api/testclient/client.go
```diff
@@ -0,0 +1,41 @@
...
+
+type RestClient struct {
+	client *addon.Client
+}
+
+//var restClient, _ = NewHubClient()
+
+// Add with and without login/token creation
+func NewHubClient() (client *addon.Client, err error) {
+	client = addon.NewClient(os.Getenv("HUB_ENDPOINT"), "")
```
jortel, line 18: Let's consider using `HUB_BASE_URL` already defined in 'settings'.

test/api/testclient/action.go
```diff
@@ -0,0 +1,15 @@
+package testclient
```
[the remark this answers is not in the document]
aufi: Good point, will rename it (maybe with more updates on the Client package).

test/api/testclient/client.go
```diff
@@ -0,0 +1,41 @@
...
+	token, err := login(client, "admin", "foobar")
+	if err != nil {
+		return
+	}
+	client.SetToken(token)
+	return
+}
+
+type LoginObject struct {
```
jortel, line 27: Why not just use api.Login instead?

test/api/testclient/client.go
jortel: Also, why not just do all of this in package `init()` instead of a new client object and functions?
See what I suggested on the issue.

test/api/application/fixtures.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+//type TestCase struct {
+//	Test        testclient.TestCase
+//	Application *api.Application
+//}
+
+//
+// Create an Application (and stop tests on failure).
+func Create(t *testing.T, application *api.Application) {
```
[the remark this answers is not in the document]
aufi: I did, partially because this becomes handy for the same resource testing (e.g.Create() before Update test and Delete() after the test). I'm likely not implementing other mathods than Create&Delete unless see a use-case.

test/api/testclient/client.go
aufi: Yes, will update it (the ```init()```method as suggested fits great for this purpose, I'm updating the (Test)Client package).

aufi:
Related to: https://github.com/konveyor/tackle2-hub/issues/262

test/api/stakeholdergroup/api_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/konveyor/tackle2-hub/api"
+	"github.com/konveyor/tackle2-hub/test/api/client"
+)
+
+func TestStakeholderGroupCRUD(t *testing.T) {
```
aufi, line 11: @jortel Would this way of naming variables etc in this test have better fit than more verbose (more different variable names depending on tested resource type) version started in application test?

test/api/client/client.go
```diff
@@ -0,0 +1,41 @@
...
+	token, err := login(client, username, password)
+	if err != nil {
+		return
+	}
+	client.SetToken(token)
+	return
+}
+
+// Login performs a login request to the hub and returns a token
+func login(client *addon.Client, username, password string) (string, error) {
```
jortel, line 34: Seems like the login can be done in init().

test/api/application/list_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"testing"
+
+	"github.com/konveyor/tackle2-hub/api"
+)
+
+func TestApplicationList(t *testing.T) {
+	samples := Samples()
+	// Create.
+	for i := range samples {
+		Create(t, &samples[i]) // Need modify parent array instead of single record created by the loop (to keep created IDs).
```
aufi, line 13: Updated Create parameter to be the parent array's item to avoid ```range``` making the item copy (which is not a pointer anymore).

addon/adapter.go
```diff
@@ -157,7 +157,7 @@ type Params map[string]interface{}
 
 //
 // inject values into path.
-func (p Params) inject(path string) (s string) {
+func (p Params) Inject(path string) (s string) {
```
aufi, line 160: To be able use the ```inject``` method in tests, it was needed make it public.

aufi:
A sample with RichClient (not fully working) to consider, but it goes slightly away from the API interaction.

https://github.com/aufi/tackle2-hub/commit/51789771ddd79002b4748a4173a254f1126c48c3

aufi:
@jortel @mansam @djzager @dymurray Hey! This is a draft for Hub API tests, we had few discussion with Jeff already, but I'd like ask you for review and your thoughts. Thanks!

We can setup a call to discuss it if needed.

test/api/application/fixtures.go
```diff
@@ -0,0 +1,52 @@
...
+func Create(t *testing.T, r *api.Application) {
+	err := Client.Post(api.ApplicationsRoot, &r)
+	if err != nil {
+		t.Fatalf("Create fatal error: %v", err.Error()) // Fatal here, Error for standard test failure or failed assertion.
+	}
+}
+
+//
+// Delete the Application (and stop tests on failure).
+func Delete(t *testing.T, r *api.Application) {
```
dymurray, line 47: Recommend renaming this to `DeleteApplication`

dymurray (commented):
My suggestion:

`Samples()` function as it exists is okay, since you are returning an array of applications to be used for tests. However, it feels like we are *close* to table driven testing without actually doing it. I think a small tweak to each test can make it table-driven without too major of a change. You should be defining a set of tests at the beginning of each function, and if you want to pull from the samples to define an application for the test that is good.

See: https://github.com/konveyor/go-konveyor-tests/blob/main/developer/applications-inventory/applications_analyze_test.go#L14 how we define the test cases at the top in a table, and this can give you flexibility of defining test cases where you expect failures and things to go wrong rather than having to write a whole separate test function for it. What  do you think?

aufi:
> My suggestion:
> 
> `Samples()` function as it exists is okay, since you are returning an array of applications to be used for tests. However, it feels like we are _close_ to table driven testing without actually doing it. I think a small tweak to each test can make it table-driven without too major of a change. You should be defining a set of tests at the beginning of each function, and if you want to pull from the samples to define an application for the test that is good.
> 
> See: https://github.com/konveyor/go-konveyor-tests/blob/main/developer/applications-inventory/applications_analyze_test.go#L14 how we define the test cases at the top in a table, and this can give you flexibility of defining test cases where you expect failures and things to go wrong rather than having to write a whole separate test function for it. What do you think?

Thanks for the comment! Actually even this PR was started with the TestCases and shouldFail true/false structure (e.g. https://github.com/konveyor/tackle2-hub/pull/268/commits/82d9c32072876970983001927ae82c2cccea2cbb or https://github.com/konveyor/tackle2-hub/pull/268/commits/ae4d8d01b1ef9eb556ff57d24afa5aa0a06d4ab2). However it turned out that it doesn't have right fit for our (mostly API) tests use-case.

(WIP - going to add examples and objectives why did it in such way).

aufi:
Added richclient methods, README and sample API integration (slow) test doing an application analysis https://github.com/konveyor/tackle2-hub/pull/268/commits/d49f50de9e7b81ad18741735ccd302d79d754198#diff-fcd2bf711a00447192da2d46171f15d0bb6302397b523e0ba92ac9f61bbb8ff7R17-R88.

There are still few things open like update API tests with more table-driven instead of "samples-driven" CRUD tests (together with considering use Samples map instead of array). But as an initial version of Hub tests, I think this PR is ready for review.

addon/adapter.go
```diff
@@ -5,10 +5,11 @@ Tackle hub/addon integration.
...
 import (
+	"os"
+
 	"github.com/konveyor/controller/pkg/logging"
 	"github.com/konveyor/tackle2-hub/settings"
 	"golang.org/x/sys/unix"
-	"os"
 )
 
 var (
```
jortel, line 15: We need to decide on import ordering.  Looks like your IDE may be organizing imports using _goimports_.  revert this change so we don't have the thrashing.

addon/adapter.go
```diff
@@ -5,10 +5,11 @@ Tackle hub/addon integration.
 package addon
 
 import (
+	"os"
+
 	"github.com/konveyor/controller/pkg/logging"
 	"github.com/konveyor/tackle2-hub/settings"
 	"golang.org/x/sys/unix"
-	"os"
```
jortel, line 11: We need to decide on import ordering.  Looks like your IDE may be organizing imports using _goimports_.  revert this change so we don't have the thrashing.

test/api/client/client.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+}
+
+//
+// Create new Hub client with login.
+// Configured with environment variables HUB_BASE_URL, KEYCLOAK_ADMIN_USER, KEYCLOAK_ADMIN_PASS.
+// When using enabled auth, don't forget set AUTH_REQUIRED=1 to make Hub load Auth settings.
+func New() (client *addon.Client, err error) {
+	baseUrl := settings.Settings.Addon.Hub.URL
+	login := api.Login{User: settings.Settings.Auth.Keycloak.Admin.User, Password: settings.Settings.Auth.Keycloak.Admin.Pass}
```
jortel, line 27: When testing with auth, I don't think this is the user/pass we should be using but fine for now so long as it works.

test/api/client/client.go
```diff
@@ -0,0 +1,53 @@
...
+
+	// Setup client.
+	client = addon.NewClient(baseUrl, "")
+
+	// Disable HTTP requests retry for network-related errors to fail quickly.
+	client.Retry = 0
```
jortel, line 33: :+1:

jortel:
In each package, importing and creating local var/types really reduces code _cruft_.
Consider taking it a step further:
```
import (
	"github.com/konveyor/tackle2-hub/api"
	c "github.com/konveyor/tackle2-hub/test/api/client"
)

var (
	Client = c.Client
+	Path = c.Path
)

+ type Params = c.Params
```

Importing the client package as `c` introduces a single letter global which I found confusing at first. Why alias the import?

Perhaps:
```
import (
	"github.com/konveyor/tackle2-hub/api"
	"github.com/konveyor/tackle2-hub/test/api/client"
)

var (
	Client = client.Client
+	Path = client.Path
)

+ type Params = client.Params
```

test/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+API tests can be executed on locally running Hub in development mode (without other Konveyor components using DISCONNECTED=1).
+
+```
+$ export DISCONNECTED=1
+$ make run
+```
+
+```
+$ export HUB_ENDPOINT=http://localhost:8080
```
jortel, line 31: HUB_BASE_URL

addon/adapter.go
aufi: Thanks, I missed this unwanted change, reverting.

On the formatting tool, happy to discuss it, for me personally, I prefer "separated" imports (from goimports), but e.g. not removing empty comment lines like goimports do by default (and that's what we use quite often in go files), but I can accept whatever we agree on.

aufi:
> In each package, importing and creating local var/types really reduces code _cruft_. Consider taking it a step further:
> 
> ```
> import (
> 	"github.com/konveyor/tackle2-hub/api"
> 	c "github.com/konveyor/tackle2-hub/test/api/client"
> )
> 
> var (
> 	Client = c.Client
> +	Path = c.Path
> )
> 
> + type Params = c.Params
> ```
> 
> Importing the client package as `c` introduces a single letter global which I found confusing at first. Why alias the import?
> 
> Perhaps:
> 
> ```
> import (
> 	"github.com/konveyor/tackle2-hub/api"
> 	"github.com/konveyor/tackle2-hub/test/api/client"
> )
> 
> var (
> 	Client = client.Client
> +	Path = client.Path
> )
> 
> + type Params = client.Params
> ```

Thanks for this point! The aliased client import as "c" is originated in _test.go files where writing client.Must/client.Should as well as client.Path/client.Params looked too verbose to me and I wanted keep the client alias consistent across files.

The ideal solution from my point of view woudl be import test's client package methods as first-class citizens into a resource test package, but that either introduces code for importing each method (which looks to me a biolerplate) or is not possible in golang AFAIK.

I'm going to provide an alternative solution on this.

Makefile
```diff
@@ -112,3 +114,18 @@ ifeq (,$(wildcard $(INSTALL_TACKLE_SH)))
 	}
 endif
 	$(INSTALL_TACKLE_SH);
+
+# Run Hub test suite.
+test: test-unit test-api test-integration
```
aufi, line 119: test -> test-unit;
test-all? - TBD

.github/workflows/main.yml
```diff
@@ -37,6 +37,28 @@ jobs:
...
+
+  test-api:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '1.19'
+      - run: |
+          make run &
```
aufi, line 57: @djzager I'm not sure about this part, we might discuss it or add in follow-up PR.

test/api/application/fixtures.go
```diff
@@ -0,0 +1,67 @@
+package application
+
+import (
+	"github.com/konveyor/tackle2-hub/api"
+	"github.com/konveyor/tackle2-hub/test/api/client"
+)
+
+var (
+	// Setup Hub API client
+	Client = client.Client
```
aufi, line 10: @jortel Import aliases were removed, assertions moved to assert package. Removal of the Client variable setup leading to ```client.Client.Post(...)``` on many lines in tests did look too strange to me, so I left it as it was.

If we are strongly opinionated about removing Client variable when other non-client methods are called via their package names, I'd  consider some kind of rename client.Client to client.Api/API to be ```client.API.Post(...)``` that looks better to me.

aufi:
__test-e2e__ task seems to fail also in another jobs, there is likely a different issue in tackle install.

test/api/application/list_test.go
```diff
@@ -0,0 +1,36 @@
...
+	for i := range samples {
+		assert.Must(t, Create(&samples[i])) // Need modify parent array instead of single record created by the loop (to keep created IDs).
+	}
+
+	// Try list.
+	got := []api.Application{}
+	err := Client.Get(api.ApplicationsRoot, &got)
+	if err != nil {
+		t.Errorf("List error: %v", err.Error())
+	}
```
djzager, line 22: This looks like the kind of thing that we don't need a specific `list_test.go` to verify. By this I mean, if you update create and delete test fixtures to list and verify the application is either in (or not in) the list, then we've verified `list` is working without churning through this scenario.

test/api/application/fixtures.go
```diff
@@ -0,0 +1,67 @@
...
+				URL:    "https://github.com/konveyor/tackle-pathfinder.git",
+				Branch: "1.2.0",
+			},
+		},
+		{
+			Name: "Minimal application",
+		},
+	}
+	return
+}
```
djzager, line 32: Maybe, instead of having `Samples()` like this. We create a public function that returns a single application. For example `MinimalApp()` or `PathfinderGit()` that can be reused.

It might not be meaningful now, but we could consider a helper function that accepts a slice of applications and just creates them. I would expect this would be helpful at the start of scenarios.

dymurray: I prefer this as oppose to calling something like `Samples[0]` in places where you wish to use the Pathfinder sample.

.github/workflows/test-nightly.yml
```diff
@@ -0,0 +1,32 @@
+name: Test nightly
+
+on:
+  schedule:
+    - cron: '13 0,12 * * *' # Regulary every 12 hours
```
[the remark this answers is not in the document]
aufi: I'm OK removing nightly build from this PR, but I believe that schedule-based application-level tests should run within the upstream CI.

test/README.md
```diff
@@ -0,0 +1,69 @@
+# Hub Tests
+
+Hub tests consist of following parts:
+- Unit tests ```$ make test```
+- REST API tests ```$ make test-api```
+- Integration tests ```$ make test-integration```
+- WIP Export/import tests
+
+All tests can be executed with ```$ make test-all``` which will run all available tests.
```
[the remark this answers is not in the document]
aufi: The test-all target is still present in Makefile line 134. However its main purpose is not to be executed in github actions, but more likely locally if needed by someone (I use it locally).

test/api/application/bucket_test.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+func TestApplicationBucket(t *testing.T) {
+	// Test Facts subresource on the first sample application only.
+	application := Samples()[0]
+
+	// Create the application.
+	assert.Must(t, Create(&application))
+
+	// Bucket test TODO// with Client.BucketGet, BucketPut etc.
```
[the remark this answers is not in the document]
aufi: Yes, each application get a bucket. The bucket provides more actions which its content that are expected to be tested within this test file.

test/api/application/list_test.go
aufi: I partially agree, but since this is an API test, I'd expect explicit test for each endpoint method.

test/api/application/delete_test.go
```diff
@@ -0,0 +1,32 @@
...
+			err := Client.Delete(rPath)
+			if err != nil {
+				t.Errorf(err.Error())
+			}
+
+			// Check the it was deleted.
+			err = Client.Get(rPath, &r)
+			if err == nil {
+				t.Errorf("Exits, but should be deleted: %v", r)
+			}
```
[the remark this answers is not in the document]
aufi: The responsibility of fixtures.go methods is to perform given action (and relay the error if returned), it doesn't know if the action should or should not raise the error.
The tests know what should be the result of calling fixtures methods (or direct API Get/Post/.. methods) and the test might not want rely on method's returned error, so it might do another cal to check the real result.
This allows both positive and negative tests and I believe, this _should not_ be part of fixtures.

Makefile
```diff
@@ -112,3 +114,21 @@ ifeq (,$(wildcard $(INSTALL_TACKLE_SH)))
 	}
 endif
 	$(INSTALL_TACKLE_SH);
+
+# Run test targets always (not producing test dirs there).
+.PHONY: test test-api test-integration
+
+# Run unit tests.
+test:
+	go test -v ./auth/
```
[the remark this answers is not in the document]
aufi: Tests generaly might have special requirements to setup (like running Hub API, theoretically running keycloak or whole Konveyor) ,so I'm not sure about scanning directories for test files. To specify unit tests-only (=> no dependencies except the golang environment), I'd prefer add it to the Makefile test section for unit test execution.

test/api/application/delete_test.go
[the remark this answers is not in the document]
djzager: This comment makes me think we are in disagreement on the purpose of these fixtures. If you would like to update the `Create()` fixture to accept an `err` that represents the error that would be returned, that would be great. But the idea of the fixtures is to strengthen the verification for these core, always repeating actions, ie. everytime we create an application we make sure 1) we can GET it 2) we can LIST it 3) it's bucket is created as expected. This way, when some addon writes a test using these fixtures these verifications are happening for free.

It's important for fixtures to verify the thing that was supposed to happen happened.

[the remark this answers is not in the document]
aufi: Probably the disagreement comes from different understanding what should be tested with different kinds of tests. The GET/LIST/BUCKET after Create makes sense to me for the API endpoint test, but e.g. the addon analyze test might not focus on the application creation action if it is bulletproof OK or not, the addon test should create the application (ideally by fixtures method) and if the Create method has not failed, use the application in its test.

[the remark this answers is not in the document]
aufi: I'm not against adding a helper which will check the created/deleted/.. resource like application, I just don't see it as a mandatory part of the fixtures.go methods.

.github/workflows/test-nightly.yml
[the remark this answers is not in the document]
aufi: Keep for now to see how it works

test/api/application/bucket_test.go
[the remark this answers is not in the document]
aufi: :+1: check the bucket object ref

test/api/application/list_test.go
aufi: Will be updated merging create with get/list.

test/api/application/fixtures.go
aufi: Used Map for fixtures for now. Will check if other options fit better in followup tests development.

aufi:
The PR was updated in last days, I'm about continue this work in folloup PRs, but would be great get this PR in (or request change if there is something not acceptable, thanks).

test/api/client/client.go
```diff
@@ -0,0 +1,50 @@
...
+	return
+}
+
+//
+// Params mapping.
+type Params map[string]interface{}
+
+//
+// Merge path with params.
+func Path(base string, paramsMap map[string]interface{}) string {
```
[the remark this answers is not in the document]
djzager: Is there a technical reason you say this?

> test files should not import addon package

test/api/application/bucket_test.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/konveyor/tackle2-hub/test/api/client"
+	"github.com/konveyor/tackle2-hub/test/assert"
+)
+
+func TestApplicationBucket(t *testing.T) {
+	// Test Facts subresource on the first sample application only.
+	application := Samples()["Minimal"]
+
+	// Create the application.
+	assert.Must(t, Create(&application))
```
djzager, line 16: So, here we use the helper function, but in the create_get_delete_test.go we use the raw client. Why?

test/api/client/client.go
```diff
@@ -0,0 +1,53 @@
...
+
+var Client *addon.Client
+
+func init() {
+	var err error
+	Client, err = New()
+	if err != nil {
+		panic(fmt.Sprintf("Error: Cannot setup API client: %v.", err.Error()))
+	}
+}
```
djzager, line 19: I think we should leave the instantiation of the client to the consumer.

test/api/client/client.go
djzager, line 27: Does this work if auth is disabled?

djzager: I assume it must since I don't see any configuring of auth in the nightly you included.

test/api/task/api_test.go
```diff
@@ -0,0 +1,88 @@
...
+		})
+	}
+}
+
+func TestTaskList(t *testing.T) {
+	samples := Samples()
+
+	for name := range samples {
+		sample := samples[name]
+		assert.Must(t, Create(&sample))
```
djzager, line 72: It's really confusing that we have the addon client, the test client, and these helper methods all being used and it's not really clear when we use which.

djzager (changes requested):
Thank you for being patient with me. I'm beginning to wonder if this PR would have been easier to review if we had just started with CRUD on applications :sunglasses: 

After further review, there are a few things I struggle with more than  the rest:

1. What are the rules for when we are using the addon.Client, the internal Client, and the helper functions? Have we considered, especially given that we aren't going to do additional verification beyond did it return 200, just using the addon.Client directly? If we are going to keep the helper functions around, they should at least  grab the object ID off the response.
2. I'm still against the Samples() `map[string]api.Application`.
3. In the application test files, we have a lot of churn on creating and deleting applications. What's preventing us from creating a batch of applications, verifying for all of them their bucket was created and we can GET it, add some facts to an application, update another application, ..., profit.

test/api/client/client.go
aufi: Yes, works with disabled auth.

test/api/application/fixtures.go
```diff
@@ -0,0 +1,60 @@
...
+			Repository: &api.Repository{
+				Kind:   "git",
+				URL:    "https://github.com/konveyor/tackle-pathfinder.git",
+				Branch: "1.2.0",
+			},
+		},
+
+	}
+	return
+}
```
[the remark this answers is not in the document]
aufi: I agree the ```application.MinimalApplication``` (or shorter  ```application.Minimal``` ) looks better. Downsides are
- there is no iterator over applications out of the box
- it is not possible modify applications before using it in the test, e.g. some time ago, test paralelization was mention, even it is not required now, the Samples() function have an ability modify e.g. Application names before returning it (like if app Name is not empty, append space+5 random chars to Name)

test/api/task/api_test.go
aufi: I'm going to update README mentioning usage of both addon and rich client methods.

test/api/client/client.go
aufi: Idea behind this is that since test packages (test/application, test/tag, etc.) provide methods for interaction with Hub API, it should be responsible for the client setup. And having this same code in each resource type test package would be a duplication, it is easier to set it up here directly within the client pkg.

test/integration/applications-inventory/analysis/basic_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/konveyor/tackle2-hub/test/api/application"
+	"github.com/konveyor/tackle2-hub/test/api/client"
+	"github.com/konveyor/tackle2-hub/test/api/task"
+	"github.com/konveyor/tackle2-hub/test/assert"
+)
+
+//
+// Test application analysis
+// "Basic" means that there no other dependencies than the application itself (no need prepare credentials, proxy, etc)
+func TestBasicAnalysis(t *testing.T) {
```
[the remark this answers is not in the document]
aufi: IIRC, it was accepted was to begin with this test as part of Hub repo. Later or with more complicated integration/e2e test cases, do it in separate repo.

test/api/application/bucket_test.go
aufi: Focus in this test file was on the buket get, so it looked to me easier to use short command to create a application for the bucket test. (this is not part of the reuse test pkg, it is its usage within the API test, so it is expected that test author might decide what fits better for its use case, more desciption in part of https://github.com/konveyor/tackle2-hub/pull/268#issuecomment-1520072758)

test/api/application/create_get_delete_test.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/konveyor/tackle2-hub/test/api/client"
+	"github.com/konveyor/tackle2-hub/test/assert"
+)
+
+func TestApplicationCreateGetDelete(t *testing.T) {
+	samples := Samples()
+	// Create on array of Applications calls subtest
+	for _, r := range samples {
+		t.Run(r.Name, func(t *testing.T) {
+			err := Client.Post(api.ApplicationsRoot, &r)
```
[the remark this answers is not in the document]
aufi: It looked to me to be more descriptive to use a command with path as parameter to be clearly visible what path is being tested.
 (this is not part of the reuse test pkg, it is its usage within the API test, so it is expected that test author might decide what fits better for its use case, more desciption in part of https://github.com/konveyor/tackle2-hub/pull/268#issuecomment-1520072758)

aufi:
@jortel, @djzager PR was updated based on yesterdays discussion - last 2 commits - details below. I'd like ask for you ack/comments to continue work this way.

A ```/binding``` package was added in following steps:
- copy-pasted client.go and error.go from addon
- modified addon adapter.go to binding richclient.go
- added  Application methods to binding/richclient based on test methods
- TODO: add other resources methods only ack-ed the way how application is done

```/test/api/application/...``` test files were updated to not use ```test/client``` (which will be removed before this PR goes in), but use the ```binding/richclient``` (+addon client from the richclient for few calls not yet supported by the richclient).

A question I'd like to confirm:
Existing addon package methods on resources structs (like Application) use ```func (h *Application) Get(id uint) (r *api.Application, err error)``` (https://github.com/konveyor/tackle2-hub/blob/main/addon/application.go#L16-L23) functions that initiate and _return_ the application.
However the addon's "raw" client returns values by modifying parameter passed as reference and the same do tests so far ```func (h *Application) Get(r *api.Application) (err error)``` (https://github.com/konveyor/tackle2-hub/pull/268/files#diff-d1be7262de63443ed32ecff32f7e38074f4c9569417ea354683fe4ec05704dd0R21-R24).
The "reference" way makes easy use of methods like ```assert.Should(t, <action>)```) that return only the error and code looks better for my point of view. A downside is little awkward way of Get method (https://github.com/konveyor/tackle2-hub/pull/268/files#diff-0eaa5b22ad02c862b6bd490ac8274d8bc5b1985949d6e3f193b9deabf1b33001R18), that needs to prepare struct where the response should be stored to and need to set its ID before doing the API call. And incompatibility with existing addon methods.

The question is: does the ```binding``` package need to implement addon's "helper" methods signature (initiate internally resources and return it) instead of using passed refs (like addon client and in tests now)?

File: binding/richclient.go [hunk 33]
```diff
@@ -0,0 +1,117 @@
...
+)
+
+func init() {
+	err := Settings.Load()
+	if err != nil {
+		panic(err)
+	}
+
+	RichClient = newRichClient()
+}
```
jortel, line 30: I think we decided to not provide the `RichClient` global.
Please remove the `var` and the `init()`.
Please rename `Adapter` => `RichClient`.

binding/richclient.go
```diff
@@ -0,0 +1,117 @@
...
+	client *Client
+}
+
+// Client provides the REST client.
+func (h *Adapter) Client() *Client {
+	return h.client
+}
+
+// newRichClient builds a new Addon RichClient object.
+func newRichClient() (richClient *Adapter) {
```
jortel, line 62: Let's rename _newRichClient()_ to just `New(url) (client *RichClient)`.
This will be the _main_ constructor for the package.

Add `RichClient.Login(user, password string) (err error)`

```
client := binding.New("http://localhost:8080/hub")
err := client.Login("elmer", "fudd")
...
```

aufi: :+1: Updated.

binding/richclient.go
aufi: :+1: Updated (the init() was left for Settings load)

jortel:
The binding should follow the same signature pattern as the _addon_ binding.
Example:
- Create(r *Application) (err error)
- Get(int id) (r *Application, err error)
- List() (list []api.Application, err error)
- Update(r *Application) (err error)
- Delete(int id) (err error)

In the future, the List() should also support _filter_.

binding/application.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"github.com/konveyor/tackle2-hub/api"
+)
+
+//
+// Application API.
+type Application struct {
+	// hub API client.
+	client *Client
+}
```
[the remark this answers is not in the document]
aufi: Ack, will update binding pkg&tests.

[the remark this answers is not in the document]
aufi: Application methods in binding package were updated according to comment above.

aufi:
The PR was updated based on discussion from call we had last week with @dymurray @jortel @djzager @mansam and clarification from Jeff this week. The PR is up for review.

I'm about submitting follow-up PRs focusing on 
- covering most of resources in basic API test with the binding package update
- expansion of Application analysis integration test
 
Once this PR gets in as a foundation for Konveyor upstream tests, I'm happy discuss other close-future tasks on tests as well as running these tests on PRs/nightly.

jortel (approved):
LGTM
Thanks for all of the hard work on this!

test/integration/applications-inventory/analysis/windup_basic_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/konveyor/tackle2-hub/api"
+	"github.com/konveyor/tackle2-hub/test/api/application"
+	"github.com/konveyor/tackle2-hub/test/assert"
+)
+
+//
+// Test application analysis
+// "Basic" means that there no other dependencies than the application itself (no need prepare credentials, proxy, etc)
+func TestBasicAnalysis(t *testing.T) {
```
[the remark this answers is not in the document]
aufi: I partially agree, but having this in test/integration directory (instead of test/api), expecting its move to different repo (or at least be just the integration test proving that Hub can triger addon analysis) and having more advanced analysis test in different repos looked OK-ish for me. Also it could be a helpful example for other&QE how to use the test methods from Hub.
Fot that reason I put it to this PR. If this would block this PR from getting in, I can remove it.

aufi:
Thanks for your feedback! Merging.

PR #270, opened 2023-03-21
:bug: Fixes transaction issues.

jortel:
Fix DB transaction issues.
1. The `TxHandler` begins a transaction using the _main_ DB client.
2. The Transaction (new client) acquires a DB connection (limited to 1) which is unknown to the _main_ DB client.
3. The Task (auth) validator attempts to validate the task ID using the _main_ DB client which BLOCKS.

Proposed Solution:

1. Remove the `TxHandler` from the main router handler chain.
5. Add a handler in the main router chain that just sets the (main) DB client in the context.
6. Add the new `Transaction` handler in (each) routeGroup (as needed) after the `auth.Required` handler which will execute after the auth handler but before each of the _route_ handlers.

Note: It seemed unnecessary to pass the k8s client through the gin.Context.


api/application.go
```diff
@@ -49,12 +49,13 @@ func (h ApplicationHandler) AddRoutes(e *gin.Engine) {
 	routeGroup.DELETE(ApplicationsRoot, h.DeleteList)
 	routeGroup.DELETE(ApplicationRoot, h.Delete)
 	// Tags
+	routeGroup = e.Group("/")
 	routeGroup.Use(auth.Required("applications"))
```
jortel, line 53: **hint**: As a result of the missing:
```
routeGroup = e.Group("/")
```
The 
```
auth.Required("applications")
```
Added the `Required` AFTER the `Transaction`.

mansam: Great job finding that, what a nightmare to debug. What do you think about maybe adding a safeguard to the top of the Transaction handler to catch situations like this? Gin exposes the names of functions in the handler chain to us, we could check to see that the Transaction handler is second to last and abort if it isn't to avoid a deadlock:

```go
func Transaction(ctx *gin.Context) {
	names := ctx.HandlerNames()
	parts := strings.Split(names[len(names) - 2], "/")
	if parts[len(parts) - 1] != "api.Transaction" {
		_ = ctx.Error(liberr.New("invalid handler chain", "chain", names))
		ctx.Abort()
		return
	}
...
```

mansam (approved):
LGTM

PR #271, opened 2023-03-22
:ghost: Skip Pathfinder calls when disconnected

aufi:
Updating application delete action that calls Pathfinder API to not do that when running as Disconnected (e.g. for fast API tests).

Otherwise gets ```$ curl -X DELETE http://localhost:8080/applications/1
{"error":"Delete \"/pathfinder/assessments/bulkDelete\": unsupported protocol scheme \"\""}```

Follow-up on https://github.com/konveyor/tackle2-hub/pull/242

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -200,7 +200,6 @@ func (h ApplicationHandler) Delete(ctx *gin.Context) {
 		_ = ctx.Error(err)
 		return
 	}
-
 	result = h.DB(ctx).Delete(m)
 	if result.Error != nil {
 		_ = ctx.Error(result.Error)


jortel (changes requested):
Let's deal with this in pathfinder.go `DeleteAssessment()`.
```
//
// DeleteAssessment deletes associated assessments by application Ids.
func (r *Pathfinder) DeleteAssessment(ids []uint, ctx *gin.Context) (err error) {
if Settings.Disconnected {
		return
	}
```

aufi:
> Let's deal with this in pathfinder.go `DeleteAssessment()`.

Good point, updated, thanks!

jortel (approved):
LGTM

PR #278, opened 2023-03-27
:ghost: Attempt to reconcile keycloak indefinitely

mansam:
Instead of trying to log in to Keycloak every second for three minutes before giving up and allowing reconcile to fail, retry every three seconds forever until logging into keycloak succeeds.


jortel (approved):
LGTM

aufi (approved):

PR #280, opened 2023-03-27
:ghost: auth.Validator must use DB passed in the gin context.

jortel:
The auth.Request enhanced to include a `*gom.DB` to ensure the `auth` package uses the DB (or Tx) passed in the gin context instead of the global (main) DB client.  This protects against deadlock caused when the global (main) DB client is used after a transaction has started.


mansam (approved):
LGTM

PR #281, opened 2023-03-28
:bug: Adjusted addon binding retry.

jortel:
PROBLEM:

The addon (binding) is designed to detect and deal with temporary hub unavailability. The primary use case is a hub pod restart. It is intended to retry requests for 10 min.  In reality, this ends up being 20 min (2 x 10).  This is because in the adapter Run(), a caught error/panic is reported to the hub which results in entering the retry logic a 2nd time.

SOLUTION:

Maintain the error state on the client so subsequent requests will fail immediately.

### DIFF
--- a/addon/adapter.go
+++ b/addon/adapter.go
@@ -80,7 +80,9 @@ func (h *Adapter) Run(addon func() error) {
 			if _, soft := err.(interface{ Soft() *SoftError }); !soft {
 				Log.Error(err, "Addon failed.")
 			}
-			h.Failed(err.Error())
+			if h.client.Error == nil {
+				h.Failed(err.Error())
+			}
 			os.Exit(1)
 		}
 	}()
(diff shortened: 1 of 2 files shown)


addon/adapter.go
```diff
@@ -80,7 +80,9 @@ func (h *Adapter) Run(addon func() error) {
 			if _, soft := err.(interface{ Soft() *SoftError }); !soft {
 				Log.Error(err, "Addon failed.")
 			}
-			h.Failed(err.Error())
+			if h.client.Error == nil {
+				h.Failed(err.Error())
```
jortel, line 84: **hint:** Don't report the error with the client (to the hub) when the client has an error because the client cannot talk to the hub.

addon/adapter.go
```diff
@@ -109,10 +111,7 @@ func (h *Adapter) Client() *Client {
 func newAdapter() (adapter *Adapter) {
 	//
 	// Build REST client.
-	client := &Client{
-		baseURL: Settings.Addon.Hub.URL,
-		token:   Settings.Addon.Hub.Token,
-	}
+	client := NewClient(Settings.Addon.Hub.URL, Settings.Addon.Hub.Token)
```
jortel, line 114: **hint**: should be using the constructor.

addon/client.go
```diff
@@ -43,23 +48,35 @@ type Client struct {
 	token string
 	// transport
 	transport http.RoundTripper
+	// Retry limit.
+	Retry int
```
jortel, line 52: **hint**: Better for the client retries configurable.

aufi (approved):
LGTM

PR #283, opened 2023-04-05
:bug: Add Auth Require to Import API

aufi:
CreateUser field was empty for ImportSummaries, it turned out that there is missing Auth middleware for Import-related API endpoints.

Adding the Required("imports") middleware to the Imports endpoint.

API returns non-empty user field now:
```
$ curl -F "file=@import.csv" http://localhost:8080/importsummaries/upload
{"id":16,"createUser":"admin.noauth","updateUser":"","createTime":"2023-04-04T18:46:45.588499423+02:00","filename":"","importStatus":"Completed","importTime":"2023-04-04T18:46:45.588499423+02:00","validCount":0,"invalidCount":0,"createEntities":true}
```

Fixes: https://issues.redhat.com/browse/MTA-129

### DIFF
--- a/api/import.go
+++ b/api/import.go
@@ -46,6 +46,7 @@ type ImportHandler struct {
 //
 // AddRoutes adds routes.
 func (h ImportHandler) AddRoutes(e *gin.Engine) {
+	e.Use(Required("imports"), Transaction)
 	e.GET(SummariesRoot, h.ListSummaries)
 	e.GET(SummariesRoot+"/", h.ListSummaries)
 	e.GET(SummaryRoot, h.GetSummary)
--- a/api/import.go
+++ b/api/import.go
@@ -47,7 +47,7 @@ type ImportHandler struct {
 // AddRoutes adds routes.
 func (h ImportHandler) AddRoutes(e *gin.Engine) {
 	routeGroup := e.Group("/")
-	routeGroup.Use(Required("imports"), Transaction)
+	routeGroup.Use(Required("applications"))
 	routeGroup.GET(SummariesRoot, h.ListSummaries)
 	routeGroup.GET(SummariesRoot+"/", h.ListSummaries)
 	routeGroup.GET(SummaryRoot, h.GetSummary)
--- a/api/import.go
+++ b/api/import.go
@@ -46,17 +46,18 @@ type ImportHandler struct {
 //
 // AddRoutes adds routes.
 func (h ImportHandler) AddRoutes(e *gin.Engine) {
-	e.Use(Required("imports"), Transaction)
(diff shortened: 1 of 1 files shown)


api/import.go
```diff
@@ -46,6 +46,7 @@ type ImportHandler struct {
 //
 // AddRoutes adds routes.
 func (h ImportHandler) AddRoutes(e *gin.Engine) {
+	e.Use(Required("imports"), Transaction)
```
mansam, line 49: You'll need to create a RouteGroup before applying the middleware, otherwise it will be applied to the entire router.

```go
routeGroup := e.Group("/")
routeGroup.Use(Required("imports"), Transaction)
routeGroup.Get(SummariesRoot, h.ListSummaries)
...
```

mansam (changes requested):

api/import.go
aufi: Thanks Sam! Updated.

mansam (approved):
LGTM

api/import.go
```diff
@@ -46,16 +46,18 @@ type ImportHandler struct {
...
-	e.GET(SummaryRoot, h.GetSummary)
-	e.DELETE(SummaryRoot, h.DeleteSummary)
-	e.GET(ImportsRoot, h.ListImports)
-	e.GET(ImportsRoot+"/", h.ListImports)
-	e.GET(ImportRoot, h.GetImport)
-	e.DELETE(ImportRoot, h.DeleteImport)
-	e.GET(DownloadRoot, h.DownloadCSV)
-	e.POST(UploadRoot, h.UploadCSV)
+	routeGroup := e.Group("/")
+	routeGroup.Use(Required("imports"), Transaction)
```
jortel, line 50: Recommend Require("applications") instread of imports.  The applications scopes already exist and mapped to users.
Also, the new `Transaction` will not be on the mta-6.1 branch.

mansam: The "imports" scope exists as well and is already mapped, but we could consolidate it with the "applications" scope. Good catch on the branch compatibility issue.

aufi: I was thinking about the application since this endpoint main purpose is to create applications (but found out it exists, so used it), will update to "applications". Good catch on Transaction!

(potential deletion of "imports" scope might be part of different PR)

jortel (approved):
LGTM

api/import.go
jortel: Good catch, Sam!

api/import.go
jortel: I didn't check if _imports_ existed and wanted to avoid creating it.
Let's keep it as _imports_.  Sorry for the confusion.

Issue #266, opened 2023-03-15
Store (new) analysis report as first-class entities.


jortel:
In support of:
- https://github.com/konveyor/enhancements/pull/111
- https://github.com/konveyor/enhancements/tree/ef8da3e53e3af719c15334a9c583a1d99d6632c1/enhancements/tackle/analyzers-report-format

Store (new) analysis report as first-class entities.
Each application may have _many_ analysis reports.
Each report will contain:
- (technology) dependencies
- Issues (violations)

Though the tags/categories are reported by the analyzer, they will be maintained in the inventory using the existing model/API.

Violations are reported by the analyzer as an array of [RuleSet](https://github.com/konveyor/analyzer-lsp/blob/main/hubapi/violations.go).

High level deliverables:
- Add models
- Add APIs.

Notes:
- Previous reports will be stripped down to only include enough data to support statistics reporting.

---

## API

Note: analys**es** is plural.

**RuleReport**:
- ruleset
- rule
- name
- description
- category
- effort
- labels []string
- applications (int)  // number of affected applications

**IssueReport**:
- id
- ruleset
- rule
- name
- description
- category
- effort
- incidents int
- files int
- application
  - id
  - name
  - effort 

**DepReport**:
- Name
- Version
- Type
- SHA
- applications (int)

**Dependency**:
- Name
- Version
- Type
- SHA
- Indirect

#### /applications/1/analyses
  - methods: GET, POST
  - return: []Report 
 
#### /applications/1/analysis
  - methods: GET
  - return: Report (Latest)
 
#### /applications/1/analysis/dependencies
  - methods: GET
  - filter:
    - direct|indirect
    - name (match: like) Example: "java.io.*"
 - return: []Dependency 

#### /applications/1/analysis/issues
  - methods: GET
  - filter: 
     - category (match: any of)
       - Example: ?filter=category=(mandatory|warning)
     - effort ((match: any of) 
       - Example: ?filter=effort=10
  - return: []Issue

#### /applications/1/analysis/report/category 
  - _FUTURE_ 
  - methods: GET 
  - filter:
  - return (count, story points collated by category)

#### /applications/1/analysis/report/effort 
  - _FUTURE_ 
  - methods: GET 
  - filter:
     - category= (default: _mandatory_)
  - return:  (count, story points of mandatory issues collated by effort)

---

#### /analyses/:id
  - methods: GET, DELETE
  - return Report 

#### /analyses/issues
  - methods: GET 
  - filter:
[...]

PR #290, opened 2023-04-07
Analysis report.

jortel:
Add models/APIs for analysis _reports_.
Adds _filter_ support.

Adds X-Total response header to support pagination in the UI.

Todo: 
- [x] Add analysis models.
- [x] Add application /analysis API sub-resource.
- [x] Add analyses/composite/
- [x] Add /analyses/issues/
- [x] Add /analyses/dependencies
- [x] Add filtering.

Iteration of: https://github.com/konveyor/tackle2-hub/issues/266

### DIFF
--- a/api/analysis.go
+++ b/api/analysis.go
@@ -480,7 +480,6 @@ func (h AnalysisHandler) IssueComposites(ctx *gin.Context) {
 	// Build.
 	ruleSets := h.rulesetIDs(ctx, &filter)
 	q := h.DB(ctx)
-	q = p.Paginated(q)
 	q = q.Select(
 		"i.RuleID",
 		"i.Category",
@@ -517,6 +516,8 @@ func (h AnalysisHandler) IssueComposites(ctx *gin.Context) {
 	// Find.
 	type M struct {
 		RuleID   string
+		Category string
+		Effort   int
 		Labels   model.JSON
 		Name     string
 		Version  string
@@ -533,9 +534,12 @@ func (h AnalysisHandler) IssueComposites(ctx *gin.Context) {
 		r := &list[i]
 		affected[r.RuleID] = r.Affected
 	}
+	q = p.Paginated(q)
 	db = h.DB(ctx)
 	db = db.Select(
 		"i.RuleID",
+		"i.Category",
+		"i.Effort",
 		"i.Labels",
 		"t.Name",
 		"t.Version",
@@ -563,10 +567,12 @@ func (h AnalysisHandler) IssueComposites(ctx *gin.Context) {
 		if !found {
 			r = &IssueComposite{
 				Affected: affected[m.RuleID],
+				Category: m.Category,
 				RuleID:   m.RuleID,
 			}
 			collated[m.RuleID] = r
 			resources = append(resources, r)
+			r.Effort += m.Effort
 			if m.Labels != nil {
 				_ = json.Unmarshal(m.Labels, &r.Labels)
 			}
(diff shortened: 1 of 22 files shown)


api/proxy.go
```diff
@@ -101,6 +101,7 @@ func (h ProxyHandler) Create(ctx *gin.Context) {
 	proxy := &Proxy{}
 	err := h.Bind(ctx, proxy)
 	if err != nil {
+		_ = ctx.Error(err)
 		return
```
jortel, line 105: **note**: unrelated, just fixed it because I noticed it.

api/analysis.go
```diff
@@ -0,0 +1,927 @@
...
+			"  AnalysisIssue issue,",
+			"  AnalysisRuleSet ruleset,",
+			"  Analysis an,",
+			"  Application app ",
+			"WHERE",
+			"  app.ID = an.ApplicationID",
+			"  AND an.ID = ruleset.AnalysisID",
+			"  AND ruleset.ID = issue.RuleSetID",
+			"  AND an.ID IN (?)",
+		},
```
jortel, line 642: **hint**: This WHERE could likely be narrowed a little more by applying the issue filter but decided not to for the first iteration out of simplicity.  Not sure it's worth it.

api/analysis.go
```diff
@@ -0,0 +1,927 @@
...
+		return
+	}
+	for i := range list {
+		m := &list[i]
+		mp[m.RuleID] = append(
+			mp[m.RuleID],
+			Ref{ID: m.AppID, Name: m.AppName})
+	}
+	return
+}
```
jortel, line 658: **hint**: This WHERE ^^ could likely be narrowed a little more by applying the issue filter but decided not to for the first iteration out of simplicity.  Even for 100k items in the result set, the overhead should be relatively small. Not sure it's worth it.

api/base.go
```diff
@@ -62,6 +63,33 @@ func (h *BaseHandler) Sorted(ctx *gin.Context) (db *gorm.DB) {
...
+			return
+		}
+	}
+	mp := ctx.Writer.Header()
+	mp[Total] = []string{
+		strconv.Itoa(int(count)),
+	}
+	return
+}
+
```
jortel, line 92: This is needed ^ to prevent caller from making a request that will return 500K+ resources and blowing up the hub.

api/bucket.go
```diff
@@ -28,12 +28,6 @@ const (
 	BucketContentRoot = BucketRoot + "/*" + Wildcard
 )
 
-//
-// Params
-const (
-	Filter = "filter"
```
jortel, line 34: **note**: Moved to pkg.go which imports from _filter/_.

api/analysis.go
mansam: Seems fine to go without for now for the sake of simplicity.

api/base.go
mansam: 500 seems like plenty to be requesting at once.

api/analysis.go
```diff
@@ -0,0 +1,1156 @@
...
+
+//
+// Routes
+const (
+	AnalysesRoot               = "/analyses"
+	AnalysisRoot               = AnalysesRoot + "/:" + ID
+	AnalysesDepsRoot           = AnalysesRoot + "/dependencies"
+	AnalysesIssuesRoot         = AnalysesRoot + "/issues"
+	AnalysesCompositeRoot      = AnalysesRoot + "/composite"
+	AnalysisCompoSiteDepRoot   = AnalysesCompositeRoot + "/dependencies"
```
mansam, line 24: typo: AnalysisCompoSiteDepRoot -> AnalysisCompositeDepRoot

mansam (approved):
Looks great!

PR #291, opened 2023-04-07
:sparkles: TagCategory API improvements

mansam:
* Adds optional `name` query parameter to `/tagcategories` list.
* Adds `/tagcategories/:id/tags` subresource, with optional `name` query parameter to filter.

These `name` query parameters are a case sensitive exact match, which will result in the list containing either one or no entries.


addon/tag.go
```diff
@@ -59,7 +59,8 @@ func (h *Tag) Delete(r *api.Tag) (err error) {
 // Find by name and type.
 func (h *Tag) Find(name string, category uint) (r *api.Tag, found bool, err error) {
 	list := []api.Tag{}
-	err = h.client.Get(api.TagsRoot, &list)
+	path := Path(api.TagCategoryTagsRoot).Inject(Params{api.ID: r.ID})
```
jortel, line 62: Shouldn't this be?
```
path := Path(api.TagCategoryTagsRoot).Inject(Params{api.ID: category})
```

addon/tag.go
```diff
@@ -59,7 +59,8 @@ func (h *Tag) Delete(r *api.Tag) (err error) {
...
+	err = h.client.Get(path, &list, Param{Key: api.Name, Value: name})
 	if err != nil {
 		return
 	}
```
jortel, line 66: Can't the loop and search following this be replaced with:
```
if len(list) > 0 {
  found = true
  r = list[0]
}

File: addon/tag.go [hunk 3]
```diff
@@ -143,7 +144,7 @@ func (h *TagCategory) Delete(r *api.TagCategory) (err error) {
 // Find by name.
 func (h *TagCategory) Find(name string) (r *api.TagCategory, found bool, err error) {
 	list := []api.TagCategory{}
-	err = h.client.Get(api.TagCategoriesRoot, &list)
+	err = h.client.Get(api.TagCategoriesRoot, &list, Param{Key: api.Name, Value: name})
 	if err != nil {
 		return
 	}
```
jortel, line 150: Can't the loop and search following this be replaced with:
```
if len(list) > 0 {
  found = true
  r = &list[0]
}

aufi (approved):

PR #293, opened 2023-04-12
:sparkles: Add support: pagination, sorting, YAML bindings.

jortel:
POST/PUT/PATCH body format _bind_ based on Content-Type (header) supporting both JSON and YAML.
GET body format rendered based on Accept (header) supporting both JSON and YAML. Defaults to JSON. Other MIMEs specified will be coerced to JSON.

Improves support for `Accept` header matching.

[1] https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

### DIFF
--- a/addon/client.go
+++ b/addon/client.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/gin-gonic/gin/binding"
 	liberr "github.com/konveyor/controller/pkg/error"
 	"github.com/konveyor/tackle2-hub/api"
 	"io"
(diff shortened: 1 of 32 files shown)


mansam (approved):
This is awesome. LGTM.

addon/client.go
```diff
@@ -22,12 +23,6 @@ import (
 	"time"
 )
 
-const (
-	Accept   = api.Accept
-	AppJson  = api.AppJson
-	AppOctet = api.AppOctet
-)
```
jortel, line 29: hint: Use MIMEs in the binding package instead.

PR #297, opened 2023-04-14
:bug: Fix accepted.

jortel:
Fixes a regression introduced when Accepted() was _enhanced_ to support wildcards.
When used to determine of bucket and file endpoints are to return file content both curl and browsers passed */* which results in the endpoints returning html instead of streaming a file.


mansam (approved):
LGTM

Issue #299, opened 2023-04-24
Jira integration: Enable insecure communication field required


gildub:
Enhancement document for Jira integration : https://github.com/konveyor/enhancements/pull/85/files#diff-cbcdb511847ad0f355f71680ac0d8ccd22fd6f1792fe2b09c8d6c512970b82e4R32
requires a "Enable insecure communication" field for Jira instances.
It seems there is no such field in the current data model for trackers to support it.

jortel:
Done.

PR #303, opened 2023-04-26
:sparkles: Support insecure connections to Jira

mansam:
Adds an `insecure` boolean field to the Tracker object which is used to determine whether to disable certificate verification for connections to Jira.

Implements https://github.com/konveyor/tackle2-hub/issues/299


jortel (approved):
LGTM

aufi (approved):

PR #307, opened 2023-05-02
Upgraded k8s

jortel:
Upgrades to controller-runtime 1.13.1 which requires Go 1.17 and k8s packages like client-go 1.25.
The latest is 1.14.1 which requires Go 1.20 and k8s packages like client-go 1.27.

Removes dep on konveyor/controller which is pinned to very old versions of _everything_ which was causing go-mod conflicts in the _new_ tackle2-addon-analyzer based on the analyzer-lsp.


mansam (approved):
LGTM.

PR #309, opened 2023-05-04
:sparkles: Add BatchHandler

mansam:
* Introduces /batch/tickets and /batch/tags endpoints
  which support POSTing a slice of the respective resources
  to create them. Creation is all or nothing-- if any fail
  to be created the entire operation is rolled back.
* Responsibility for rendering responses has been delegated
  to a Render handler. The Render handler builds the response
  from values set on a Response struct on the rich context.
* Moves the k8s Client into the rich context


api/context.go
```diff
@@ -50,6 +80,7 @@ func Transaction(ctx *gin.Context) {
 			rtx.DB = db
 			if len(ctx.Errors) > 0 {
 				err = ctx.Errors[0]
+				ctx.Errors = nil
```
mansam, line 83: Clearing this fixes the error duplication that was seen on endpoints using the transaction handler.

jortel: Ah.  good find.

jortel (approved):
LGTM

PR #318, opened 2023-05-10
:bug: Improve Jira issue status reporting

mansam:
* report status based on status category rather than status name
* Make sure tracker error message is cleared when a connection is reestablished


jortel (approved):
LGTM
Unrelated but noticed nothing has docstirngs in this file.

PR #322, opened 2023-05-11
:bug: Update Swagger fix array/single, add analysis

aufi:
QE reported failures on Get actions for jobfunction, group. Updating swagger annotation to expect single value instead of array (jobfunction and review) and make a small fix when Get returned model instead of api package struct (group).

Also adding Analysis endpoint methods to generated swagger spec.

Fixes: https://issues.redhat.com/browse/MTA-582

### DIFF
--- a/api/analysis.go
+++ b/api/analysis.go
@@ -119,7 +119,7 @@ func (h AnalysisHandler) AppLatest(ctx *gin.Context) {
 // @description Resources do not include relations.
 // @tags analyses
 // @produce json
-// @success 200 {object} []api.Analyses
+// @success 200 {object} []api.Analysis
 // @router /analyses [get]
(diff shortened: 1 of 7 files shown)


api/analysis.go
```diff
@@ -119,7 +119,7 @@ func (h AnalysisHandler) AppLatest(ctx *gin.Context) {
 // @description Resources do not include relations.
 // @tags analyses
 // @produce json
-// @success 200 {object} []api.Analyses
+// @success 200 {object} []api.Analysis
```
aufi, line 122: Understood that this method is an action on multiple Analyses, but we don't have defined that as special type, so swagger failed to re-generate its specs. Modified it to an "array of analysis" ```[]api.Analysis``` (same in other changes in analysis.go file).

jortel (approved):
Thanks @aufi !

PR #326, opened 2023-05-11
:bug: Fix possible nil ptr dereference in Jira refresh

mansam:
It's possible for the Jira client to return a nil response, so we need to check it before using it. Other locations are already safe because the `handleJiraError` function checks the response for nil.


jortel (approved):
LGTM

PR #329, opened 2023-05-15
:sparkles: Add rules to rulesets.

jortel:
The _new_ analyzer ruleset (rules) can be split up into files.

rules/
  |__ruleset.yaml
  |__rules1.yaml
  |__rules2.yaml


jortel:
Taking another approach.

PR #331, opened 2023-05-16
:seedling: Add Identity and Proxy to API tests.

aufi:
Adding Indetity and Proxy resources to API tests. Identity provides samples with login/password, key and settings file.

Related to https://github.com/konveyor/tackle2-hub/pull/268

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,15 +123,15 @@ endif
 
 # Run unit tests.
 test:
-	go test -v ./auth/
+	go test -count=1 -v ./auth/
 
 # Run Hub REST API tests.
 test-api:
-	HUB_BASE_URL=${HUB_BASE_URL} go test -v ./test/api/...
+	HUB_BASE_URL=${HUB_BASE_URL} go test -count=1 -v ./test/api/...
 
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -123,15 +123,15 @@ endif
 
 # Run unit tests.
 test:
-	go test -v ./auth/
+	go test -count=1 -v ./auth/
 
 # Run Hub REST API tests.
 test-api:
-	HUB_BASE_URL=${HUB_BASE_URL} go test -v ./test/api/...
+	HUB_BASE_URL=${HUB_BASE_URL} go test -count=1 -v ./test/api/...
```
aufi, line 130: ```-count=1``` disabled go test caching that is On by default and if code was not changes between test executions, tests are not executed, but result provided from cache (like ```PASS
ok  	github.com/konveyor/tackle2-hub/test/api/identity	(cached)
```), this is not wanted e.g. in case tests execution can be affected with environment variables like ```HUB_BASE_URL```.

jortel (approved):
LGTM

PR #341, opened 2023-05-19
:ghost: Add swagger spec change notification for QE

aufi:
Related to https://issues.redhat.com/browse/MTA-615

### DIFF
--- a/.github/workflows/swagger-change-notification.yml
+++ b/.github/workflows/swagger-change-notification.yml
@@ -16,4 +16,4 @@ jobs:
         uses: thollander/actions-comment-pull-request@v2
         with:
           message: |
-          :mega: There is a Swagger OpenAPI spec files change, please consider re-generate your API client. @mguetta1 @sshveta @kpunwatk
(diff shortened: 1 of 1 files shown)


.github/workflows/swagger-change-notification.yml
```diff
@@ -0,0 +1,19 @@
...
+    name: Notify QE on Swagger spec file change
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Comment PR
+        uses: thollander/actions-comment-pull-request@v2
+        with:
+          message: |
+          :mega: There is a Swagger OpenAPI spec files change, please consider re-generate your API client. @mguetta1 @sshveta @kpunwatk
```
jortel, line 19: @djzager Do you think we should we be hard coding user names in github actions?

aufi: :+1: Would you mind create such team, it looks I don't have permissions to do that.

djzager: Looks like we already have @konveyor/tackle-qe

djzager (approved):
lgtm

PR #342, opened 2023-05-19
:ghost: Add code snip to hack analysis.

jortel:
Adds a code snip for the first 5 incidents.
Makes the number of generated incidents configurable.
Makes each the issue name unique by ruleset and rule.


mturley (approved):
LGTM

PR #345, opened 2023-05-22
Test swagger update - do not merge

aufi:
Test to see if comment with QE mention will appear.


aufi:
Getting error ```Error: Resource not accessible by integration```, that is probably caused by insufficent permissions for the github action. That can be adjusted at https://github.com/konveyor/tackle2-hub/settings/actions (bottom of the page Workflow / Read write permission; but I don't have such privileges).

PR #349, opened 2023-05-25
:sparkles: Add analysis total effort and include in the IssueReport.

jortel:
Add analysis total effort and include in the IssueReport.

Also, noticed the IssueReport is not marshalled with `Issue` inline in YAML.  This is because the yaml:,inline tag is missing. However, the report redefines the incidents field.  This (field override) works with the json decoder but not the yaml decoder.

### DIFF
--- a/api/analysis.go
+++ b/api/analysis.go
@@ -235,6 +235,7 @@ func (h AnalysisHandler) AppCreate(ctx *gin.Context) {
 			_ = ctx.Error(err)
 			return
 		}
+		analysis.Effort += r.Effort * len(r.Incidents)
 	}
 	//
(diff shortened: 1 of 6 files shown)


api/analysis.go
```diff
@@ -1184,7 +1200,6 @@ func (h *AnalysisHandler) withLabels(m interface{}, ctx *gin.Context, f *qf.Filt
 //
 // AnalysisManifest EST resource.
 type AnalysisManifest struct {
-	Resource     `yaml:",inline"`
```
jortel, line 1187: hint: not relevant.

mansam (approved):
LGTM

PR #350, opened 2023-05-25
:ghost: Add test badges to README

aufi:
Adding on-merge Hub build workflow and nightly tests badges to README (both are green as of now).


aufi:
Screenshot:
image

PR #356, opened 2023-05-30
:ghost: Using multi-part form instead of manifest.

jortel:
The approach of: using `File` and `AnalysisManifest` is intended to be more RESTful but after test driving it seems overly complicated.  Seems more straight forward to use a multi-part for as the _Manifest_.

Adds addon binding Client.FileSend() used by the `File` and `Application.Analysis` APIs.

### DIFF
--- a/api/base.go
+++ b/api/base.go
@@ -246,7 +246,7 @@ func (h *BaseHandler) Bind(ctx *gin.Context, r interface{}) (err error) {
 }
 
 //
-// Decoder returns a decoder based on Content-Type header.
+// Decoder returns a decoder based on encoding.
 // Opinionated towards json.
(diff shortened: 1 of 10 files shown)


api/base.go
```diff
@@ -248,13 +248,14 @@ func (h *BaseHandler) Bind(ctx *gin.Context, r interface{}) (err error) {
 //
 // Decoder returns a decoder based on Content-Type header.
```
mansam, line 249: This docstring will need to be updated

mansam (approved):
LGTM

mansam (approved):
The update to use the content type on the form data looks great.

aufi (approved):

PR #358, opened 2023-05-30
:sparkles: Sort validation.

jortel:
Revert the usage of h.Paginated().
Removed BaseHandler.Paginated() because - Paginated endpoints need to fetch the count using the un-paginated DB client.  Then, add the pagination and sorting when fetching the models. 

Moved the sorting object to the _new_ sort package.

The Sort.With() enhanced to provide both validation (of supported files) and mapping to actual selected columns.

### DIFF
--- a/api/analysis.go
+++ b/api/analysis.go
@@ -125,10 +125,10 @@ func (h AnalysisHandler) AppList(ctx *gin.Context) {
 	resources := []Analysis{}
 	// Build query.
 	id := h.pk(ctx)
-	db := h.Paginated(ctx)
+	db := h.DB(ctx)
 	db = db.Where("ApplicationID = ?", id)
-	count := int64(0)
 	// Count.
+	count := int64(0)
 	result := db.Model(&model.Analysis{}).Count(&count)
 	if result.Error != nil {
 		_ = ctx.Error(result.Error)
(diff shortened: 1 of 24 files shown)


api/analysis.go
```diff
@@ -345,14 +346,27 @@ func (h AnalysisHandler) AppDeps(ctx *gin.Context) {
 			{Field: "id", Kind: qf.LITERAL},
 			{Field: "name", Kind: qf.STRING},
 			{Field: "version", Kind: qf.STRING},
-			{Field: "type", Kind: qf.STRING},
```
jortel, line 348: **hint**: `type` no longer exists.

api/analysis.go
```diff
@@ -345,14 +346,27 @@ func (h AnalysisHandler) AppDeps(ctx *gin.Context) {
...
 			{Field: "indirect", Kind: qf.STRING},
+			{Field: "labels", Kind: qf.STRING},
```
jortel, line 350: **hint**: `Labels` missing.

api/analysis.go
```diff
@@ -394,7 +409,8 @@ func (h AnalysisHandler) AppDeps(ctx *gin.Context) {
 // @description List application issues.
 // @description filters:
 // @description - id
-// @description - ruleid
```
jortel, line 397: **hint**: ruleid replaced by ruleset, rule.

PR #359, opened 2023-05-31
:sparkles: Omit Ref empty Name in API

aufi:
There is a Ref API struct used e.g. in Tag-TagCategory association. API e.g. on Tag create doesn't load the Category so the API reponse doesn't contain Category Name.

Trying to solve this **either** with pre-loading the Category on TagCreate or drop name field from the Ref API response. **To be discussed**, looking forward to comments in this PR.

Related to https://issues.redhat.com/browse/MTA-674

### DIFF
--- a/api/base.go
+++ b/api/base.go
@@ -374,10 +374,10 @@ func (r *Resource) nameOf(m interface{}) (name string) {
 //
 // Ref represents a FK.
 // Contains the PK and (name) natural key.
-// The name is read-only.
+// The name is optional and read-only.
 type Ref struct {
 	ID   uint   `json:"id" binding:"required"`
-	Name string `json:"name"`
+	Name string `json:"name,omitempty"`
 }
 
 //
--- a/api/tag.go
+++ b/api/tag.go
@@ -98,6 +98,8 @@ func (h TagHandler) Create(ctx *gin.Context) {
 		return
 	}
 	m := r.Model()
+	// Load associated TagCategory.
+	h.DB(ctx).First(&m.Category, m.CategoryID)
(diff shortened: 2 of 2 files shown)


api/base.go
```diff
@@ -374,10 +374,10 @@ func (r *Resource) nameOf(m interface{}) (name string) {
 //
 // Ref represents a FK.
 // Contains the PK and (name) natural key.
-// The name is read-only.
+// The name is optional and read-only.
 type Ref struct {
 	ID   uint   `json:"id" binding:"required"`
-	Name string `json:"name"`
+	Name string `json:"name,omitempty"`
```
jortel, line 380: This seems like a good holistic change.

api/tag.go
```diff
@@ -98,6 +98,8 @@ func (h TagHandler) Create(ctx *gin.Context) {
 		return
 	}
 	m := r.Model()
+	// Load associated TagCategory.
+	h.DB(ctx).First(&m.Category, m.CategoryID)
```
jortel, line 102: There are likely many places where resources are created with refs.  Further we could (and probably do) create resources with nested resources containing refs.  This is likely a _can of worms_ to be avoid.
For now, I wonder if we can **not** include this change and set the expectation that POST operations will only return resources with nameless refs?

aufi: Agreed on can of worms setting Name on all such places. Keeping Ref with ID mandatory (as until now), but setting right expectation, that the Name really don't have to be provided (so will be omitted) sounds good to me too!

jortel (approved):
LGTM

PR #363, opened 2023-05-31
Fix Pathfinder assessment counting

mansam:
The UI makes POST requests to /pathfinder/assessments/assessment-risk in order to show the calculated risk on the side bar. This needs to be excluded in order to be sure we're only counting initiated assessments.

Check that POST requests are to /pathfinder/assessments.

Fixes https://issues.redhat.com/browse/MTA-682


jortel (approved):
LGTM

Issue #370, opened 2023-06-05
Add API test for Dependency resource


aufi:
There is a Dependency resource which allows linking applications to dependency trees (like one application depends on another, e.g. web depends on backend).

Recomended steps:
- Look on existing API tests
- Create dependency package/directory within test/api (https://github.com/konveyor/tackle2-hub/tree/main/test/api)
- Create meaningful sample test data
- Add needed methods to binding package (methods in new file for dependency, https://github.com/konveyor/tackle2-hub/tree/main/binding)
- Create basic test creating dependency linking at least 2 applications
- Check the test is passing, update API test matrix and submit PR!

For more information check https://github.com/konveyor/tackle2-hub/tree/main/test or https://www.youtube.com/watch?v=Ax6rFADyXDo&t=70s

Part of https://github.com/konveyor/tackle2-operator/issues/220 for @khareyash05

aufi:
@khareyash05 Nice work on basic Dependency test from #375! Please add a follow-up PR extending the Dependency API test with:
- add test to ensure there are no cyclic dependencies
- - direct: when Gateway->Inventory dependency exists, API shoud not allow Inventory->Gateway dependency to be created
- - indirect: when Gateway->Inventory and Inventory->Database dependencies exist, API shoud not allow Database->Gateway dependency to be created
(just for info, related code inside the Hub is https://github.com/konveyor/tackle2-hub/blob/main/migration/v2/model/dependency.go#L30)
- update assertEqual to real check with if !assert.FlatEqual or use ==
- add green check marks to https://github.com/konveyor/tackle2-hub/pull/371/files#diff-60f6c1cb4a3382a5a4b5f570e7efe0ee215d5f1de49134b5a112ed4c7a240c64R8 (once merged)

PR #375, opened 2023-06-06
:seedling: add dependency  package in test/api

khareyash05:
Upstream issue [https://github.com/konveyor/tackle2-operator/issues/220](https://github.com/konveyor/tackle2-operator/issues/220)

Issue in the repo #370 

added api_test.go,pkg.go,samples.go in `dependency` directory. Also added `dependency.go` in binding package and added dependency client in `binding/richclient.go`

### DIFF
--- a/api/dependency.go
+++ b/api/dependency.go
@@ -64,35 +64,6 @@ func (h DependencyHandler) Get(ctx *gin.Context) {
 // @produce json
 // @success 200 {object} []api.Dependency
 // @router /dependencies [get]
-// func (h DependencyHandler) List(ctx *gin.Context) {
-// 	var list []model.Dependency
-
-// 	db := h.Paginated(ctx)
-// 	to := ctx.Query("to.id")
-// 	from := ctx.Query("from.id")
-// 	if to != "" {
-// 		db = db.Where("toid = ?", to)
-// 	} else if from != "" {
-// 		db = db.Where("fromid = ?", from)
-// 	}
-
-// 	db = h.preLoad(db, clause.Associations)
-// 	result := db.Find(&list)
-// 	if result.Error != nil {
-// 		_ = ctx.Error(result.Error)
-// 		return
-// 	}
-
-// 	resources := []Dependency{}
-// 	for i := range list {
-// 		r := Dependency{}
-// 		r.With(&list[i])
-// 		resources = append(resources, r)
-// 	}
-
-// 	h.Respond(ctx, http.StatusOK, resources)
-// }
-
 func (h DependencyHandler) List(ctx *gin.Context) {
 	var list []model.Dependency
 
--- a/api/dependency.go
+++ b/api/dependency.go
@@ -1,11 +1,10 @@
 package api
 
 import (
-	"net/http"
-
 	"github.com/gin-gonic/gin"
 	"github.com/konveyor/tackle2-hub/model"
 	"gorm.io/gorm/clause"
+	"net/http"
 )
 
 //
--- a/api/dependency.go
+++ b/api/dependency.go
@@ -68,15 +68,12 @@ func (h DependencyHandler) List(ctx *gin.Context) {
 	var list []model.Dependency
 
 	db := h.Paginated(ctx)
-	from := ctx.Query("from")
-	to := ctx.Query("to")
-
-	if from != "" && to != "" {
-		db = db.Where("fromid = ? AND toid = ?", from, to)
+	to := ctx.Query("to.id")
+	from := ctx.Query("from.id")
+	if to != "" {
+		db = db.Where("toid = ?", to)
 	} else if from != "" {
 		db = db.Where("fromid = ?", from)
-	} else if to != "" {
-		db = db.Where("toid = ?", to)
 	}
 
 	db = h.preLoad(db, clause.Associations)
--- a/api/dependency.go
+++ b/api/dependency.go
@@ -1,10 +1,11 @@
 package api
 
 import (
+	"net/http"
+
 	"github.com/gin-gonic/gin"
 	"github.com/konveyor/tackle2-hub/model"
 	"gorm.io/gorm/clause"
-	"net/http"
 )
 
 //
@@ -63,16 +64,48 @@ func (h DependencyHandler) Get(ctx *gin.Context) {
 // @produce json
 // @success 200 {object} []api.Dependency
 // @router /dependencies [get]
+// func (h DependencyHandler) List(ctx *gin.Context) {
+// 	var list []model.Dependency
+
+// 	db := h.Paginated(ctx)
+// 	to := ctx.Query("to.id")
+// 	from := ctx.Query("from.id")
+// 	if to != "" {
+// 		db = db.Where("toid = ?", to)
+// 	} else if from != "" {
+// 		db = db.Where("fromid = ?", from)
+// 	}
+
+// 	db = h.preLoad(db, clause.Associations)
+// 	result := db.Find(&list)
+// 	if result.Error != nil {
+// 		_ = ctx.Error(result.Error)
+// 		return
+// 	}
+
+// 	resources := []Dependency{}
+// 	for i := range list {
+// 		r := Dependency{}
+// 		r.With(&list[i])
+// 		resources = append(resources, r)
+// 	}
+
+// 	h.Respond(ctx, http.StatusOK, resources)
+// }
+
 func (h DependencyHandler) List(ctx *gin.Context) {
 	var list []model.Dependency
 
 	db := h.Paginated(ctx)
-	to := ctx.Query("to.id")
-	from := ctx.Query("from.id")
-	if to != "" {
-		db = db.Where("toid = ?", to)
+	from := ctx.Query("from")
+	to := ctx.Query("to")
+
+	if from != "" && to != "" {
+		db = db.Where("fromid = ? AND toid = ?", from, to)
 	} else if from != "" {
 		db = db.Where("fromid = ?", from)
+	} else if to != "" {
+		db = db.Where("toid = ?", to)
 	}
 
 	db = h.preLoad(db, clause.Associations)
--- a/api/dependency.go
+++ b/api/dependency.go
@@ -1,10 +1,11 @@
 package api
 
 import (
+	"net/http"
+
 	"github.com/gin-gonic/gin"
 	"github.com/konveyor/tackle2-hub/model"
 	"gorm.io/gorm/clause"
-	"net/http"
 )
 
 //
@@ -115,7 +116,7 @@ func (h DependencyHandler) Create(ctx *gin.Context) {
 		_ = ctx.Error(err)
 		return
 	}
-
+	r.With(m)
 	h.Respond(ctx, http.StatusCreated, r)
 }
 
--- a/test/api/dependency/api_test.go
+++ b/test/api/dependency/api_test.go
@@ -12,8 +12,8 @@ func TestDependencyCRUD(t *testing.T) {
 	for _, sample := range Samples {
 		t.Run(fmt.Sprintf("Dependency from %s -> %s", sample.ApplicationFrom.Name, sample.ApplicationTo.Name), func(t *testing.T) {
 
-			assert.Should(t, Application.Create(&sample.ApplicationFrom))
-			assert.Should(t, Application.Create(&sample.ApplicationTo))
+			assert.Must(t, Application.Create(&sample.ApplicationFrom))
+			assert.Must(t, Application.Create(&sample.ApplicationTo))
 
 			// Create.
 			dependency := api.Dependency{
@@ -38,9 +38,14 @@ func TestDependencyCRUD(t *testing.T) {
 			// Delete dependency.
 			assert.Should(t, Dependency.Delete(dependency.ID))
 
+			_, err = Dependency.Get(dependency.ID)
+			if err == nil {
+				t.Errorf("Resource exits, but should be deleted: %v", dependency)
+			}
+
 			//Delete Applications
-			assert.Should(t, Application.Delete(sample.ApplicationFrom.ID))
-			assert.Should(t, Application.Delete(sample.ApplicationTo.ID))
+			assert.Must(t, Application.Delete(sample.ApplicationFrom.ID))
+			assert.Must(t, Application.Delete(sample.ApplicationTo.ID))
 		})
 	}
 }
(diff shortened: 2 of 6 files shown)


khareyash05:
Not able to handle the Samples array as it expects `[]api.Dependency{}`

test/api/dependency/samples.go
```diff
@@ -0,0 +1,44 @@
...
+
+	testCase = TestCase{
+		Name:         "Test",
+		Dependency:   dependency,
+		Application1: aliceApplication,
+		Application2: bobApplication,
+	}
+	tc      = []TestCase{testCase}
+	Samples = []api.Dependency{tc}
+)
```
aufi, line 44: Definition of Dependencies (and likely also TestCases) will need to be updated. There is some limitation caused by api.Dependency keeping just Ref of Aplication, so we need store application elsewhere. I can see few options:
- Use TestCases struct that might even not contain the Dependency itself, but something like just ApplicationFrom and ApplicationTo, so the test will go through testcases and create ApplicationTo, ApplicationFrom and then dependency between these applications (the dependency doesn't have any special fields than the Application refs, so it should work well).
- Use api.Dependency model, that stores links to Application (not just Refs), see https://github.com/konveyor/tackle2-hub/blob/main/migration/v2/model/dependency.go#L12-L18 (yes, we do have slightly different models for database and API). The the Samples array would be array of these model.Dependency and test could ceate both Applications and "convert" the "convert" the model.Dependency to api.Dependency using method With https://github.com/konveyor/tackle2-hub/blob/main/api/dependency.go#L157-L161
- It would great if you can try both ways (or maybe find something else if you'd like it more than suggestions above) and we can discuss how it works if needed.

aufi (changes requested):
Hey Yash, thanks for your PR! There are few things we need to work on (comments in the source code), but overall looks good!

test/api/dependency/samples.go
khareyash05: I think I have some idea regarding the first option ( I shall try creating the TestCases without the dependency and then after try to inject a sample dependency in the Samples array). And pass similar structure to the `api_test.go` . An example of structure in `api.test.go` according to what I can undestand can be 
```
applicationFrom := api.Application{
				Name:        r.From.Name,
				Description: "From application",
}

applicationTo := api.Application{
			       Name:        r.To.Name,
				Description: "To application",
}
dependency := api.Dependency{
				From: api.Ref{
					ID: applicationFrom.ID,
				},
				To: api.Ref{
					ID: applicationTo.ID,
				},
}
```
and then pass like this `err := Dependency.Create(&dependency)`. Am i going towards the right direction ?

khareyash05: > Definition of Dependencies (and likely also TestCases) will need to be updated. There is some limitation caused by api.Dependency keeping just Ref of Aplication, so we need store application elsewhere. I can see few options:
> 
> * Use TestCases struct that might even not contain the Dependency itself, but something like just ApplicationFrom and ApplicationTo, so the test will go through testcases and create ApplicationTo, ApplicationFrom and then dependency between these applications (the dependency doesn't have any special fields than the Application refs, so it should work well).
> * Use model.Dependency model, that stores links to Application (not just Refs), see https://github.com/konveyor/tackle2-hub/blob/main/migration/v2/model/dependency.go#L12-L18 (yes, we do have slightly different models for database and API). The Samples array would be array of these model.Dependency and test could create both Applications and "convert" the model.Dependency to api.Dependency using method With() https://github.com/konveyor/tackle2-hub/blob/main/api/dependency.go#L157-L161
> * It would great if you can try both ways (or maybe find something else if you'd like it more than suggestions above) and we can discuss how it works if needed.

I think 1st option is easier to implement because it involves less complexities and 2nd option is is less intuitive which can be easily dealt by the 1st. What's your say on that?

test/api/dependency/api_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"testing"
+
+	"github.com/konveyor/tackle2-hub/test/assert"
+)
+
+func TestDependencyCRUD(t *testing.T) {
+	for _, r := range Samples {
+		t.Run("Dependency_CRUD", func(t *testing.T) {
+			// Create.
+			err := Dependency.Create(&r)
```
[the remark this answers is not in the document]
khareyash05: As I understood from the review that dependency need To/From fields which will be used to refer to the applications?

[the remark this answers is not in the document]
khareyash05: and that To/From will be generated by creating an instance of application in `api_test.go`. Am i in the right direction?

test/api/dependency/samples.go
khareyash05: I have a thought process of injecting dependency withdependecy-less TestCase in Samples[]. and then use From/To in the `api_test.go` to access them in the `dependency` to be used by various CRUD functions.

test/api/dependency/api_test.go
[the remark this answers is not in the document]
aufi: Yes

test/api/dependency/samples.go
aufi: I'm not sure I fully understand, but sounds good.

test/api/dependency/api_test.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"testing"
+
+	"github.com/konveyor/tackle2-hub/api"
+	"github.com/konveyor/tackle2-hub/test/assert"
+)
+
+func TestDependencyCRUD(t *testing.T) {
+	for _, r := range Samples {
+		t.Run("Dependency_CRUD", func(t *testing.T) {
```
aufi, line 12: Just a nit, but both Dependency and CRUD is part of the method name (2 lines above), so it is already part of test output, so it would be better add a new information instead. I'd suggest update to something like ApplicationFrom.Name->ApplicationTo.Name (with fmt.Sprintf).

test/api/dependency/api_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+			applicationFrom := api.Application{
```
aufi, line 14: These lines actually don't create the applications. It initialize variables, but doesn't call API (so IDs will be nil). Please update to Application.Create calls, if you pass tc.ApplicationFrom/To as references (with &) it will be updated with ID "automatically".

test/api/dependency/samples.go
```diff
@@ -0,0 +1,54 @@
...
+
+// Set of valid resources for tests and reuse.
+type TestCase struct {
+	Name            string
+	ApplicationFrom api.Application
+	ApplicationTo   api.Application
+}
+
+var (
+	dependency = api.Dependency{
```
aufi, line 15: Some of these lines seems to be redundant to me (the dependency here and init() below). But I'd suggest get it to (partially) working state (passing tests against locally running Hub https://github.com/konveyor/tackle2-hub/tree/main/test#rest-api) and we can refactor it then.

test/api/dependency/api_test.go
```diff
@@ -0,0 +1,96 @@
...
+			// Get.
+			got, err := Dependency.Get(dependency.ID)
+			if err != nil {
+				t.Errorf(err.Error())
+			}
+			if assert.FlatEqual(got, dependency.ID) {
+				t.Errorf("Different response error. Got %v, expected %v", got, dependency.ID)
+			}
+
+			_, err = Dependency.Get(r.ID)
```
aufi, line 47: Since the update was removed, this Get is likely not needed.

aufi (commented):
Thanks for update! Left few comments, let's sync tomorrow.

test/api/dependency/api_test.go
khareyash05: Missed the line! Will do that

khareyash05: shall i pass in `applicationFrom/To` in Application.Create()? will that be ok?

test/api/dependency/samples.go
khareyash05: Refactored the code Kindly review

api/dependency.go
```diff
@@ -1,10 +1,11 @@
 package api
 
 import (
+	"net/http"
+
```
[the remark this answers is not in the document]
khareyash05: I don't know it automatically changes on clicking save is there any way to stop this?

test/api/dependency/api_test.go
```diff
@@ -0,0 +1,119 @@
...
+	for _, r := range Samples {
+		t.Run(fmt.Sprintf("Dependency from %s -> %s", r.From.Name, r.To.Name), func(t *testing.T) {
+
+			applicationFrom := api.Application{
+				Name:        r.From.Name,
+				Description: "From application",
+			}
+
+			err := app.Application.Create(&applicationFrom)
+			if err != nil {
```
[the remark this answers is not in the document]
khareyash05: Shall I do it for Dependency.Create()/Get() as well apart from Applications?

khareyash05:
TODO: Cyclic Dependency.

test/api/dependency/api_test.go
```diff
@@ -0,0 +1,77 @@
...
+			// Create.
+			dependency := api.Dependency{
+				From: api.Ref{
+					ID: applicationFrom.ID,
+				},
+				To: api.Ref{
+					ID: applicationTo.ID,
+				},
+			}
+			assert.Must(t, Dependency.Create(&dependency))
```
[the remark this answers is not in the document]
khareyash05: Right Must will return a fatal error. Will take care of that in future.

test/api/dependency/api_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/konveyor/tackle2-hub/api"
+	"github.com/konveyor/tackle2-hub/test/assert"
+)
+
+func TestDependencyCRUD(t *testing.T) {
+	for _, sample := range Samples {
+		t.Run(fmt.Sprintf("Dependency from %s -> %s", sample.ApplicationFrom.Name, sample.ApplicationTo.Name), func(t *testing.T) {
+
+			applicationFrom := sample.ApplicationFrom
+			assert.Must(t, Application.Create(&applicationFrom))
```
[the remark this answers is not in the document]
khareyash05: Just used it to reduce the length of variable.Will make changes.

test/api/dependency/api_test.go
```diff
@@ -0,0 +1,70 @@
...
+			got, err := Dependency.Get(dependency.ID)
+			if err != nil {
+				t.Errorf(err.Error())
+			}
+			if assert.FlatEqual(got, dependency.ID) {
+				t.Errorf("Different response error. Got %v, expected %v", got, dependency.ID)
+			}
+
+			// Delete dependency.
+			assert.Should(t, Dependency.Delete(dependency.ID))
```
aufi, line 39: I missed it meanwhile, but after Delete, there should be a Get testing the Dependency was really deleted, example: https://github.com/konveyor/tackle2-hub/blob/main/test/api/identity/api_test.go#L48-L51

test/api/dependency/api_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+			//Delete Applications
+			assert.Should(t, Application.Delete(sample.ApplicationFrom.ID))
```
aufi, line 42: Cleanup of test data, use Must.

aufi (commented):
:+1: Needed some clarification on Must/Should and fix List test - recommend run the test locally more than once to see, there one test run doesn't leave data for another.

test/api/dependency/api_test.go
```diff
@@ -0,0 +1,70 @@
...
+		})
+	}
+}
+
+func TestDependencyList(t *testing.T) {
+	for _, sample := range Samples {
+
+		// Create applications.
+		assert.Should(t, Application.Create(&sample.ApplicationFrom))
+		assert.Should(t, Application.Create(&sample.ApplicationTo))
```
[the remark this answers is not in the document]
khareyash05: but we are deleting applicationFrom.ID as it is not needed. Kindly help me in understanding why will we need this line

khareyash05:
@aufi Kindly review

api/dependency.go
```diff
@@ -67,12 +68,15 @@ func (h DependencyHandler) List(ctx *gin.Context) {
 	var list []model.Dependency
 
 	db := h.Paginated(ctx)
-	to := ctx.Query("to.id")
-	from := ctx.Query("from.id")
-	if to != "" {
-		db = db.Where("toid = ?", to)
+	from := ctx.Query("from")
```
aufi, line 71: Why such change? Change of parameter names would break e.g. UI. A place for the change is likely the List method in binding package (add a ListFiltered(params) method which will pass the params to client.Get), but let's leave this to follow-up PR with non-basic tests like cyclic dependency test.

test/api/dependency/api_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	// List dependencies.
+	got, err := Dependency.List()
+	if err != nil {
+		t.Errorf(err.Error())
+	}
+	if assert.FlatEqual(got, Samples) {
+		t.Errorf("Different response error. Got %v, expected %v", got, Samples)
+	}
```
aufi, line 79: Found out we're using the assert.FlatEqual method in a wrong way (also in some other tests outside of this PR), the condition should look like ```if !assert.FlatEqual(got, Samples) { //error ```

test/api/dependency/api_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// Delete Dependencies.
+	for _, dependency := range got {
```
aufi, line 82: Thihs would delete all dependencies in API, not only ones created by this test. Not saying it is blocker in this case, but better track created objects to be deleted.

test/api/dependency/api_test.go
```diff
@@ -0,0 +1,91 @@
...
+		assert.Should(t, Dependency.Delete(dependency.ID))
+	}
+
+	// Delete applications.
+	for _, sample := range Samples {
+		assert.Must(t, Application.Delete(sample.ApplicationFrom.ID))
+		assert.Must(t, Application.Delete(sample.ApplicationTo.ID))
```
aufi, line 89: I'm not sure if this can pass more than once, samples array is not updated in loop starting line 54. Use variables you had before or update the samples array with like ```Samples[i] = sample``` e.g. on line 59 (this is a weird feature of golang loops).

test/api/dependency/api_test.go
khareyash05: Was wondering too when working on the part before! This might be a breaking change in the code. Have to dig deep to addressing the same.

test/api/dependency/api_test.go
khareyash05: Ya that's something should be taken care of.  A way to tackle this what I can think of is creating an array of ID's and store ID's of dependencies at their time of creation. The Delete() of the Dependency will work on the array of ID's and will only delete them.

khareyash05: As we have (or we will) ensure that all the created Dependencies are present in the Dependency.List(), we will ensure that all the dependencies are used and thus they are allowed to be deleted.

test/api/dependency/api_test.go
khareyash05: This is a little weird feature tbh! Will update the name of the variable

aufi (approved):
There are some things (assert with FlatEqual and add cyclic test), that need to be updated in followup PR, but:

Looks good to me, nice work Yash!

khareyash05:
> There are some things (assert with FlatEqual and add cyclic test), that need to be updated in followup PR, but:
> 
> Looks good to me, nice work Yash!

Thank you and will do that in the followup PR

PR #376, opened 2023-06-06
:sparkles: Seeds: Missing space in description

gildub:
Resolves https://issues.redhat.com/browse/MTA-500


jortel (changes requested):
Thanks @gildub .
Updating in the v3 migration is good because it is the canonical definition of the seed (which we need to keep tack of).  In konveyor 1.0 we have a redesign of the seeding mechanism that will make maintaining seeding much easier/better.

For upgrades, v3 will not be re-applied so this needs to be patched in v4.
Recommend adding (something like) to v4/migrate.go
```
func (r Migration) updateBundleSeed(db *gorm.DB) (err error) {
	db = db.Model(&model.RuleSet{})
	db = db.Where("Name", "Open Liberty")
	err = db.Update(
		"Description",
		"A comprehensive set of rules for migrating traditional WebSphere" +
			" applications to Open Liberty.").Error
	return
}
```
And calling at the end of Migration.Apply().

jortel (approved):
LGTM

PR #384, opened 2023-06-09
:bug: Update new TagCategory in CSV import

aufi:
Application CSV import can create new Tag Categories. For performance reasons, list of Tag Categories is loaded from database before processing all Tag imports. This list wasn't updated with newly created Tag Categories, so it was not possible create a new Tag Category and assign multiple Tags to it on the Same application.

Adding new created Tag categories to the cached Tag Categories list.

Fixes: https://issues.redhat.com/browse/MTA-480


mansam (approved):
Looks good. Do you think we should be doing the same thing with newly created tags?

mansam (approved):
LGTM

PR #387, opened 2023-06-10
:seedling: add cross dependency test

khareyash05:
Follow up PR for  #370 

- added tests for cross dependency check 
- Updated `test-api-matrix.md`

### DIFF
--- a/docs/test-api-matrix.md
+++ b/docs/test-api-matrix.md
@@ -5,7 +5,7 @@ Endpoint | binding functions | basic CRUD test | deeper test| notes/component/st
 **Application Inventory**||||
 application|:white_check_mark: partially|:heavy_check_mark:||
 bucket||||partially within application
-dependency|:white_check_mark:|:white_check_mark:|:white_check_mark:|
+dependency|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|
 file||||
 import||||
 review||||
--- a/docs/test-api-matrix.md
+++ b/docs/test-api-matrix.md
@@ -5,7 +5,7 @@ Endpoint | binding functions | basic CRUD test | deeper test| notes/component/st
 **Application Inventory**||||
 application|:white_check_mark: partially|:heavy_check_mark:||
 bucket||||partially within application
-dependency||||
+dependency|:white_check_mark:|:white_check_mark:|:white_check_mark:|
 file||||
 import||||
 review||||
--- a/test/api/dependency/api_test.go
+++ b/test/api/dependency/api_test.go
@@ -33,7 +33,7 @@ func TestDependencyCRUD(t *testing.T) {
 			if err != nil {
 				t.Errorf(err.Error())
 			}
-			if assert.FlatEqual(got, dependency.ID) {
+			if !assert.FlatEqual(got.ID, dependency.ID) {
 				t.Errorf("Different response error. Got %v, expected %v", got, dependency.ID)
 			}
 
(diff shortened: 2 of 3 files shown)


docs/test-api-matrix.md
```diff
@@ -5,7 +5,7 @@ Endpoint | binding functions | basic CRUD test | deeper test| notes/component/st
 **Application Inventory**||||
 application|:white_check_mark: partially|:heavy_check_mark:||
 bucket||||partially within application
-dependency||||
+dependency|:white_check_mark:|:white_check_mark:|:white_check_mark:|
```
aufi, line 8: Please use ```:heavy_check_mark:``` since with CRUD & cyclic check this test Dependency should be fully covered.

test/api/dependency/api_test.go
```diff
@@ -15,6 +15,8 @@ func TestDependencyCRUD(t *testing.T) {
 			assert.Must(t, Application.Create(&sample.ApplicationFrom))
 			assert.Must(t, Application.Create(&sample.ApplicationTo))
 
+			// Create a check for reverse dependency
```
aufi, line 18: Looks to be a not needed comment here.

test/api/dependency/api_test.go
```diff
@@ -101,3 +103,106 @@ func TestDependencyList(t *testing.T) {
 		assert.Must(t, Application.Delete(dependency.To.ID))
 	}
 }
+
+func TestReverseDependency(t *testing.T) {
+	for _, r := range ReverseSamples {
+		// Create Applications.
+		assert.Must(t, Application.Create(&r.Application1))
+		assert.Must(t, Application.Create(&r.Application2))
+		assert.Must(t, Application.Create(&r.Application3))
```
aufi, line 112: :+1:

test/api/dependency/api_test.go
```diff
@@ -101,3 +103,106 @@ func TestDependencyList(t *testing.T) {
...
+
+		secondDependency := api.Dependency{
+			From: api.Ref{
+				ID: r.Application2.ID,
+			},
+			To: api.Ref{
+				ID: r.Application3.ID,
+			},
+		}
+		err = Dependency.Create(&secondDependency)
```
aufi, line 135: Looks OK, but might be easier use ```assert.Should(t,Dependency.Create(&secondDependency))``` instead of condition below (when the API call should not fail).

test/api/dependency/api_test.go
```diff
@@ -101,3 +103,106 @@ func TestDependencyList(t *testing.T) {
...
+		}
+		err = Dependency.Create(&sixthDependency)
+		if err == nil {
+			t.Error("Direct Reverse dependency not allowed")
+		}
+
+		// Delete Dependencies.
+		assert.Should(t, Dependency.Delete(firstDependency.ID))
+		assert.Should(t, Dependency.Delete(secondDependency.ID))
+		assert.Should(t, Dependency.Delete(thirdDependency.ID))
```
aufi, line 198: Even API works in the way, that DELETE on not existing resources doesn't fail, just do nothing, it might be cleaner not delete dependencies that were not created.

aufi (commented):
Nice work Yash! Added few comments and one more is about naming dependencies below.

firstDependency, secondDependency, ... doesn't give much information to someone else reading the code, please modify these names e.g. adding Cyclic or Reverse keywords to ones expected to fail. (also Dependency word is probably not needed that much since all code is about dependencies, but that is just naming nit, not blocking merging).

Once updated please make it ready for review, I'll check it once more later.

test/api/dependency/api_test.go
khareyash05: Oops forgot to delete it. Was trying to add the test in the CRUD test before.

test/api/dependency/api_test.go
khareyash05: Actually did add the assert.Should condition but tried to maintain uniformity with the failing dependencies, so added this condition

test/api/dependency/api_test.go
khareyash05: Okay will remove the lines

aufi:
Related to https://github.com/konveyor/tackle2-operator/issues/220

PR #390, opened 2023-06-13
:seedling:  add csv import test

### LINKED ISSUE #388: Add CSV Application Import test
There is a feature that allows upload CSV with Applications and related resources to populate the application inventory in the Hub. Write an API test that will upload sample CSV and verify created resources in the Hub.

For most of things follow what you learned in https://github.com/konveyor/tackle2-hub/issues/370, just updating binding pakage is not needed in this case, the CSV can be uploaded with RichClient.Client HTTP methods. Suggested test structure is similar to previous PR, but expecting to specify list of TestCases (path to import CSV file and applications, dependencies, etc. expected be created) instead of samples.

Resources:
- import endpoint: https://github.com/konveyor/tackle2-hub/blob/main/api/import.go
- CSV upload method from ^ is https://github.com/konveyor/tackle2-hub/blob/main/api/import.go#L212
- sample CSV: https://github.com/konveyor/tackle2-ui/blob/main/client/public/template_application_import.csv

Example upload with curl: curl -F "file=@import.csv" http://localhost:8080/importsummaries/upload (not for test, just to show how the request should look like)

Part of https://github.com/konveyor/tackle2-operator/issues/220 for @khareyash05

khareyash05:
Closes #388 

Work Done as of now
1. Created a sample of data to be passed in `samples.go`
2. Initialized RichClient.Client
3. Uploaded CSV using FilePost
4. Get contents of `/importsummaries` 
5. Compare Applications and Dependencies with the output of the API
6. Compare valid count with no. of Applications and Dependencies
7. Added checks for all routes
8. Download the CSV
9. Compare informations of CSV's


test/api/importCSV/api_test.go
```diff
@@ -0,0 +1,29 @@
+package importCSV
```
aufi, line 1: Please name the package without capital letters, like importcsv. Check https://go.dev/doc/effective_go

test/api/importCSV/api_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"strconv"
+	"testing"
+)
+
+func TestImportCRUD(t *testing.T) {
```
aufi, line 8: CRUD is not probably the right name here.

test/api/importCSV/samples.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+type TestCase struct {
+	fileName             string
+	ExpectedApplications []api.Application
+	ExpectedDependencies []api.Dependency
+}
+
+var (
+	Applications = []api.Application{
```
aufi, line 14: Since more testcases (CSV files&expected resources to be created) are expected, I'd suggest move this directly into TestCase array.

test/api/importCSV/api_test.go
```diff
@@ -0,0 +1,29 @@
...
+	for _, r := range Samples {
+		t.Run("CSV_Import", func(t *testing.T) {
```
aufi, line 10: E.g. testcase CSV filename might be useful here instead of CSV_Import.

aufi (commented):
Nice start, Yash! Added few comments primary on naming. About extending the client.go with file upload, there is a PR https://github.com/konveyor/tackle2-hub/pull/393 updating FileSend, let's see if it is useful also for this purpose once it gets merged.

Looking forward to update checking of Imports status and created resources to find all things needed that need to be tested there.

khareyash05:
> Nice start, Yash! Added few comments primary on naming. About extending the client.go with file upload, there is a PR #393 updating FileSend, let's see if it is useful also for this purpose once it gets merged (will review then).
> 
> Looking forward to update checking of Imports status and created resources to find all things needed that need to be tested there.

Thanks Marek. Ya I went through the PR and seems to me that merging it will cause merge conflicts. Better to wait. 
As for update checking I am figuring out a few things and will update them as soon as the it is done. Thank you for the review.

khareyash05:
Added comparison logic for applications and dependencies with the expected API output

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,67 @@
...
+				t.Errorf("CSV import failed")
+			}
+			// access the import sectio of output data to checks applications and dependencies.
+			imports := outputImport["Imports"]
+
+			// fetch ExpectedApplications and ExpectedDependencies.
+			expectedApps := r.ExpectedApplications
+			expectedDeps := r.ExpectedDependencies
+			importList, ok := imports.([]interface{})
+			if ok {
```
aufi, line 34: Look to API source code, there is ImportSummary struct as well as Import struct which is a middle step between Import CSV and created Applications (https://github.com/konveyor/tackle2-hub/blob/main/api/import.go#L271-L289 and processed in https://github.com/konveyor/tackle2-hub/blob/main/importer/manager.go#L51).

Try to check Applications directly with application.List() and fields on ImportSummary like count of valid records imported. 

Generaly, it is always good to check the source code and the feature that is being tested to find out how it works, identify what might make sense to be checked (in addition to fields of an Applications/Dependencies) and since go is a static language, we should utilize its advantages, so use interfaces/casting only in necessary situations.

aufi (commented):
Good progress, added few comments.

test/api/importcsv/samples.go
```diff
@@ -0,0 +1,67 @@
...
+type TestCase struct {
+	fileName             string
+	ExpectedApplications []api.Application
+	ExpectedDependencies []api.Dependency
+}
+
+var (
+	TestCases = []TestCase{
+		{
+			fileName: "template_application_import.csv",
```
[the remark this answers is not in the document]
khareyash05: Is there any specific reference to this name change? Because I found `UploadCSV()` using `fileName`, thatswhy named the same

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/konveyor/tackle2-hub/api"
+)
+
+func TestImportCSV(t *testing.T) {
+	for _, r := range TestCases {
+		t.Run(r.FileName, func(t *testing.T) {
+
+			// Upload CSV.
+			inputData := make(map[string]interface{})
+			err := Client.FilePost("/importsummaries/upload", r.FileName, &inputData)
```
aufi, line 16: The path string should be replaced with a route const similary as in other API tests - https://github.com/konveyor/tackle2-hub/blob/main/api/import.go#L34

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,114 @@
...
+			if err != nil {
```
aufi, line 17: Minor note, the same error check would do assert.Should(), but no problem.

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,114 @@
...
+				t.Errorf(err.Error())
+			}
+
+			// Check list of Applications.
+			importedApps, _ := Application.List()
+			expectedApps := r.ExpectedApplications
+			for i, importedApp := range importedApps {
+				if i >= len(expectedApps) {
```
aufi, line 25: I'd suggest check the len() of importedApps (or likely gotApps to be consistent with other tests) before the loop  to avoid running len(expected) in each step with always the same result. Loops with applications/dependencies checks might need a refactoring.

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,114 @@
...
+				}
+				expectedApp := expectedApps[i].Name
+				if importedApp.Name != expectedApp {
+					t.Errorf("Mismatch in imported Application: Expected %s, Actual %s", expectedApp, importedApp.Name)
+				}
+			}
+
+			// Check list of Dependencies.
+			importedDeps, _ := Dependency.List()
+			expectedDeps := r.ExpectedDependencies
```
aufi, line 37: Are these expectedDeps&expectedApps needed, wouldn't be better use r.Expected... variables?

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,114 @@
...
+					continue
+				}
+				expectedDep := expectedDeps[i].To.Name
+				if importedDep.To.Name != expectedDep {
+					t.Errorf("Mismatch in imported Application: Expected %s, Actual %s", expectedDep, importedDep.To.Name)
+				}
+			}
+
+			// fetch id of CSV file and convert it into required formats
+			id := uint64(inputData["id"].(float64))
```
aufi, line 50: No need to deal with interfaces, few lines below you use ```api.ImportSummary```, use it on lines 15,16 too, then you'll get correct types out-of-the-box

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,114 @@
...
+			}
+
+			// Delete summary of the Input ID.
+			err = Client.Delete("/importsummaries/" + inputID)
+			if err != nil {
+				t.Errorf("CSV delete failed")
+			}
+
+			// Delete related Applications.
+			err = Application.Delete(uint(id))
```
aufi, line 102: This ID variable doesn't belong to the Application or Dependency, I don't think the Delete part of the tests works (cleanup of importsummary, applications and dependencies).

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,114 @@
...
+			}
+
+			// Get summaries of the Input ID.
+			outputImportSummary := api.ImportSummary{}
+			err = Client.Get("/importsummaries/"+inputID, &outputImportSummary)
+			if err != nil {
+				t.Errorf("Could not get the CSV output")
+			}
+
+			// Get all imports.
```
aufi, line 75: I can understand this and following lines do an API requests, but get importsummary response is used around line 55 - good. On imports - when doing requests, please check if the response contains relevant data (e.g. an application) also inputID seems to me to be importsummary ID, Import should have its own ID.

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+			// Check list of Applications.
+			importedApps, _ := Application.List()
+			expectedApps := r.ExpectedApplications
+			for i, importedApp := range importedApps {
+				if i >= len(expectedApps) {
+					t.Errorf("Extra imported Application: %s", importedApp.Name)
+					continue
+				}
+				expectedApp := expectedApps[i].Name
```
aufi, line 29: The Name check on application is not enough, please check also other imported fields (if matches to expected values described in testcase/sample).

aufi (changes requested):
Thanks for update! See comments inline + try follow similar conventions as in previous dependency test, e.g. when creating paths for PAI request, use consts from API package instead of hardcoding paths in string (details inline).

binding/client.go
```diff
@@ -134,7 +135,7 @@ func (r *Client) Get(path string, object interface{}, params ...Param) (err erro
 			err = liberr.Wrap(err)
 			return
 		}
-		err = json.Unmarshal(body, object)
+		err = json.Unmarshal(body, &object)
```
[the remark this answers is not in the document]
khareyash05: I had been testing a few things earlier. Forgot to revert the line.Sorry for the same

test/api/importcsv/api_test.go
khareyash05: Will update it to assert.Should() to make the tests similar in nature

test/api/importcsv/api_test.go
khareyash05: Does that also imply to /imports/ route as well

test/api/importcsv/api_test.go
khareyash05: Will do that.

test/api/importcsv/samples.go
```diff
@@ -0,0 +1,67 @@
...
+			ExpectedApplications: []api.Application{
```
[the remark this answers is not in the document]
djzager: So the data in this CSV is supposed to be equivalent to what is being described in the `template_application_import.csv`?

test/api/importcsv/api_test.go
khareyash05: Will do that! Also converted equality checks with assert.FlatEqual()

test/api/importcsv/samples.go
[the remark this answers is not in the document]
khareyash05: > So the data in this CSV is supposed to be equivalent to what is being described in the `template_application_import.csv`?

Ya even I think so because we are checking that if all the entities are being uploaded or not thus have to check them all

binding/client.go
```diff
@@ -368,7 +369,7 @@ func (r *Client) BucketPut(source, destination string) (err error) {
 			if isDir {
 				err = r.putDir(part, source)
 			} else {
-				err = r.putFile(part, source)
+				err = r.loadFile(part, source)
```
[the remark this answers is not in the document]
aufi: Understood, but this change came in time when there was another function calling the loadFile, which is gone since you added FilePost method (which is a better solution for this). So the putFile _is_ called only from BucketPut, so I'd would not recommend changing it (as actually unrelated code to this test/PR).

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,78 @@
...
+			assert.Must(t, Client.FilePost(api.SummariesRoot+"/upload", r.FileName, &inputData))
+
+			// Check list of Applications.
+			gotApps, _ := Application.List()
+			expectedApps := r.ExpectedApplications
+			if len(gotApps) != len(expectedApps) {
+				t.Errorf("Mismatch in number of imported Applications: Expected %d, Actual %d", len(expectedApps), len(gotApps))
+			} else {
+				for i, importedApp := range gotApps {
+					assert.FlatEqual(expectedApps[i].Name, importedApp.Name)
```
aufi, line 26: These assertions return true/false, but actually do nothing with the test result. Please look to the FlatEqual method body and to your previous PR to its usage (e.g. https://github.com/konveyor/tackle2-hub/blob/main/test/api/dependency/api_test.go#L34-L36)

In some cases, it might be better use == instead of FlatEqual function (expectedApps[i].Name, importedApp.Name => if expectedApps[i].Name != importedApp.Name {error...})

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/konveyor/tackle2-hub/test/assert"
+)
+
+func TestImportCSV(t *testing.T) {
+	for _, r := range TestCases {
+		t.Run(r.FileName, func(t *testing.T) {
+
+			// Upload CSV.
+			inputData := api.ImportSummary{}
+			assert.Must(t, Client.FilePost(api.SummariesRoot+"/upload", r.FileName, &inputData))
```
aufi, line 17: This is not the right way creating the path, please follow what you had in your previous PR https://github.com/konveyor/tackle2-hub/pull/375/files#diff-6ded4e5e60480daa4db7cf0740809d6b470d937e85d1b674b02777560fed1a38R25

test/api/importcsv/api_test.go
khareyash05: Will do that!

migration/v2/model/dependency.go
```diff
@@ -11,9 +11,9 @@ var depMutex sync.Mutex
 
 type Dependency struct {
 	Model
-	ToID   uint         `gorm:"index"`
+	ToID   uint         `gorm:"uniqueIndex:dependency_primary_key"`
 	To     *Application `gorm:"foreignKey:ToID;constraint:OnDelete:CASCADE"`
-	FromID uint         `gorm:"index"`
+	FromID uint         `gorm:"uniqueIndex:dependency_primary_key"`
```
aufi, line 16: Appreciate your effort on looking thi issue, but it cannot be in this PR with tests. Please create a new one referencing to reported issue and look on notes below.
- the schema is versioned, it is needed create new version (or modify the latest one if it was not released yet - I'm not sure if it is possible now, let's see on the new PR)
- I'm not sure about using "primary_key" as part of name of composite uniq key, would you consider a more descriptive name?

khareyash05: Surely ! Will address it in the next PR

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,221 @@
...
+				t.Errorf(err.Error())
+			}
+
+			// Compare number of records present.
+			if len(item1) != len(item2) {
+				t.Errorf("Mismatch in number of records present")
+			}
+
+			// Compare each value.
+			for i := 0; i < len(item1); i++ {
```
aufi, line 191: Happy to see such precise file content diff check (byte by byte), but it might be easier/more matching to the test/better to read both files data, cast []byte to string and compare it and in case of error, print both contents got and expected, that will be much easier to debug why a test potentialy failed.

binding/client.go
```diff
@@ -21,7 +21,7 @@ import (
 	pathlib "path"
 	"path/filepath"
 	"strings"
-	"time"
+	"time"
```
aufi, line 24: Nit, but not needed change, please remove trailing spaces.

docs/test-api-matrix.md
```diff
@@ -7,7 +7,7 @@ application|:white_check_mark: partially|:heavy_check_mark:||
 bucket||||partially within application
 dependency|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|
 file|:heavy_check_mark:|:heavy_check_mark:||
-import||||
+import|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|
```
aufi, line 10: The binding part is not covered for import endpoint, which is OK, but don't add :heavy_check_mark: in the first column, leave it empty or add minus symbol or something similar.

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,166 @@
...
+			// Upload CSV.
+			inputData := api.ImportSummary{}
+			assert.Must(t, Client.FilePost(api.UploadRoot, r.FileName, &inputData))
+
+			// Since uploading the CSV happens asynchronously we need to wait for the upload to check Applications and Dependencies.
+			time.Sleep(time.Second * 3)
+
+			// Check list of Applications.
+			gotApps, _ := Application.List()
+			expectedApps := r.ExpectedApplications
```
aufi, line 28: Not needed declaration of expectedApps, why don't use r.ExpectedApplications directly? (commented the same thing few lines below few days back..)

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,166 @@
...
+func TestImportCSV(t *testing.T) {
+	for _, r := range TestCases {
+		t.Run(r.FileName, func(t *testing.T) {
+
+			// Upload CSV.
+			inputData := api.ImportSummary{}
+			assert.Must(t, Client.FilePost(api.UploadRoot, r.FileName, &inputData))
+
+			// Since uploading the CSV happens asynchronously we need to wait for the upload to check Applications and Dependencies.
+			time.Sleep(time.Second * 3)
```
aufi, line 24: I think we discussed a loop waiting for importsummary until it get process all records (e.g. sleep 1 sec, get importsummary and if valid+invalid count is equal to all application&dependencies count, break from the loop and continue).

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,166 @@
...
+			// Check list of Dependencies.
+			gotDeps, _ := Dependency.List()
+			expectedDeps := r.ExpectedDependencies
+			if len(gotDeps) != len(expectedDeps) {
+				t.Errorf("Mismatch in number of imported Dependencies: Expected %d, Actual %d", len(expectedDeps), len(gotDeps))
+			} else {
+				for i, importedDep := range gotDeps {
+					expectedDep := expectedDeps[i].To.Name
+					if importedDep.To.Name != expectedDep {
+						t.Errorf("Mismatch in imported Dependency: Expected %s, Actual %s", expectedDep, importedDep.To.Name)
```
aufi, line 68: Is there a reason to not check also ```From``` side of the dependency? Consider adding it.

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,166 @@
...
+					}
+				}
+			}
+
+			// fetch id of CSV file and convert it into required formats
+			var inputID = strconv.FormatUint(uint64(inputData.ID), 10) // to be used for API compatibility
```
aufi, line 74: I don't see a reason doing this for ImportSummary, why not use ```inputData.ID``` directly on line 87
```pathForImportSummary := binding.Path(api.SummaryRoot).Inject(binding.Params{api.ID: inputData.ID})```? Ping me if there is some weird go error that requires this casting.

File: test/api/importcsv/api_test.go [hunk 28]
```diff
@@ -0,0 +1,166 @@
...
+
+			var outputImportSummaries []api.ImportSummary
+			outputMatchingSummary := api.ImportSummary{}
+			assert.Should(t, Client.Get(api.SummariesRoot, &outputImportSummaries))
+			for _, imp := range outputImportSummaries {
+				if uint(imp.ID) == inputData.ID {
+					outputMatchingSummary = imp
+				}
+			}
+			assert.FlatEqual(len(expectedApps)+len(expectedDeps), outputMatchingSummary.ValidCount)
```
aufi, line 84: Already commented the same thing some lines above in this file. ```assert.FlatEqual``` returns true if arguments are equal, false if different - that doesn't fail the test. Use it in condition like:
```if !assert.FlatEqual(...) { t.Error(error_results_are_different) }```

The advantage of assert.FlatEqual is that it checks multiple attributes on struct types variables. For basic types line string or int, it might be shorter use just == operator in the condition.

File: test/api/importcsv/api_test.go [hunk 29]
```diff
@@ -0,0 +1,166 @@
...
+						t.Errorf("Mismatch in name of import: Expected %s, Actual %s", expectedApps[j].Name, imp["applicationName"])
+					}
+					if expectedApps[j].Description != imp["description"] {
+						t.Errorf("Mismatch in name of import: Expected %s, Actual %s", expectedApps[j].Description, imp["description"])
+					}
+					if expectedApps[j].BusinessService.Name != imp["businessService"] {
+						t.Errorf("Mismatch in name of import: Expected %s, Actual %s", expectedApps[j].BusinessService.Name, imp["businessService"])
+					}
+					j++
+				}
```
aufi, line 110: IIUC, shouldn't there be a second part of the condition - if outputImports has more records than expectedApps, the test should fail too. Same for dependencies few lines below.

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,166 @@
...
+			// Delete import summary
+			assert.Should(t, Client.Delete(pathForImportSummary))
+
+			// Delete all imports
+			id := 1
+			for id <= len(expectedApps)+len(expectedDeps) {
+				pathForImport := binding.Path(api.ImportRoot).Inject(binding.Params{api.ID: id})
+				assert.Should(t, Client.Delete(pathForImport))
+				id++
+			}
```
aufi, line 163: Delete also imported Applications.

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,166 @@
...
+			err = os.Remove(pathToGotCSV)
+			if err != nil {
+				t.Errorf(err.Error())
+			}
+
+			// Delete import summary
+			assert.Should(t, Client.Delete(pathForImportSummary))
+
+			// Delete all imports
+			id := 1
```
aufi, line 158: This is not the right way - expecting ID of objects there are created by client (not seeded by the application). I'd suggest skip the Imports deletetion and be OK with ImportSummary delete (that should take imports too).

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,166 @@
...
+						t.Errorf("Mismatch in name of import: Expected %s, Actual %s", expectedDeps[k].From.Name, imp["applicationName"])
+					}
+					if expectedDeps[k].To.Name != imp["dependency"] {
+						t.Errorf("Mismatch in name of import: Expected %s, Actual %s", expectedDeps[k].To.Name, imp["dependency"])
+					}
+					k++
+				}
+			}
+
+			// Download the csv.
```
aufi, line 123: Not a blocker, the Download CSV test part looks OK, but I don't think it should be part of the import test since it actually doesn't import anything and is kind of independent on most of code in this test method. Would you mind move it to separate test method like TestDownloadCSV?

aufi (changes requested):
Thanks for update Yash! Added few comments, please look on them.

test/api/importcsv/api_test.go
aufi: This is not a blocker, just for info.

docs/test-api-matrix.md
khareyash05: I am leaving it empty

test/api/importcsv/api_test.go
khareyash05: Sorry just because it was a long name thatswhy

test/api/importcsv/api_test.go
khareyash05: Sure

test/api/importcsv/api_test.go
aufi: Not a problem, keep it as it is.

test/api/importcsv/api_test.go
khareyash05: outputimports consist of both applications and dependencies thus shouldn't be equal to length of sums?

docs/test-api-matrix.md
```diff
@@ -7,7 +7,7 @@ application|:white_check_mark: partially|:heavy_check_mark:||
 bucket||||partially within application
 dependency|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|
 file|:heavy_check_mark:|:heavy_check_mark:||
-import||||
+import||:heavy_check_mark:|:heavy_check_mark:|
```
aufi, line 10: A small update, please remove also "deeper test" (the second check mark), we'd need add also "negative" test covering rejected records etc. for a deeper test (but that don't have to be in this PR).

test/api/importcsv/api_test.go
```diff
@@ -0,0 +1,171 @@
...
+				}
+				time.Sleep(time.Second)
+			}
+
+			// Check list of Applications.
+			gotApps, _ := Application.List()
+			if len(gotApps) != len(r.ExpectedApplications) {
+				t.Errorf("Mismatch in number of imported Applications: Expected %d, Actual %d", len(r.ExpectedApplications), len(gotApps))
+			} else {
+				for i, importedApp := range gotApps {
```
aufi, line 47: This is a tiny thing, not requiring change, but there as a ```gotApps``` array - looks OK and in the loop is ```importedApp```, works, sure, but wouldn't ````gotApp`` match better to the ```gotApps```?

test/api/importcsv/api_test.go
aufi: If  the CSV for this test would get one more line with an application, but the test (expected applications) was not changed (=> more imports type 1, but not changed len(expectedApps)), it looks to me to pass this part of the test anywa - that was the thing I didn't like. On the other hand there is a check of gotApps len() on line 44 and in this PR we don't have to do deeper rejection tests, so it is likely OK.

aufi (approved):
Left few minor comments, but looks good to me.

test/api/importcsv/api_test.go
khareyash05: Will do the change

aufi:
Part of https://github.com/konveyor/operator/issues/220

PR #393, opened 2023-06-13
:ghost: Consolidated REST client.

jortel:
Consolidate the low level REST clients in binidng and addon packages.  Replaces binding.Client with addon.Client.

The addon/Client had already been updated to:
- use pipes to eliminate holding uploaded/downloaded files in memory.
- adds FileSend for to support multi-file upload.


aufi (approved):
LGTM

PR #396, opened 2023-06-14
:warning: Update Fact API to use qualified keys (source:fact)

mansam:
Get/replace/delete an anonymous fact
GET/PUT/DELETE /application/1/facts/myfact

Get/replace/delete a qualified fact
GET/PUT/DELETE /application/1/facts/mysource:myotherfact

List and replace for anonymous source:
GET/PUT /application/1/facts

List and replace all by source
GET/PUT /application/1/facts/mysource:


api/application.go
```diff
@@ -616,8 +602,13 @@ func (h ApplicationHandler) FactGet(ctx *gin.Context) {
...
-	source := ctx.Param(Source)[1:]
+
+	source, key := qualifiedFact(ctx.Param(Key))
+	if key == "" {
+		h.FactList(ctx, source)
+		return
+	}
+
 	list := []model.Fact{}
 	result = h.DB(ctx).Find(&list, "ApplicationID = ? AND Key = ? AND source = ?", id, key, source)
```
jortel, line 613: This is fine but generally we have been doing this in other places to reduce the raw sql.
```
db = h.DB(ctx)
db = db.Where("ApplicationID", id)
db = db.Where("Source", source)
db = db.Where("Key", key)
```

api/application.go
```diff
@@ -678,33 +669,35 @@ func (h ApplicationHandler) FactCreate(ctx *gin.Context) {
...
+		return
+	}
+
 	result = h.DB(ctx).First(&model.Fact{}, "ApplicationID = ? AND Key = ? AND source = ?", id, key, source)
 	if result.Error == nil {
 		result = h.DB(ctx).
 			Model(&model.Fact{}).
 			Where("ApplicationID = ? AND Key = ? AND source = ?", id, key, source).
-			Update("Value", value)
+			Update("Value", f.Value)
```
jortel, line 700: Since the model is being fetched (and it's super small), seems like it would be simpler to:
```
m := &model.Fact{}
db.First(m)
m.Value = f.Value
db.Save(m)
```

jortel:
I wonder if it would be good to _codify_ the fact key format as an object to be shared by the _api_, _binding_, and _addon_ packages.

Something like:

```
type FactKey string

func (r *FactKey) With(source, name string) {
   *r = FactKey(
      strings.Join(
         []string{source, name},
         ":"))
}

func (r *FactKey) Source() (s string) {
   if x, _, found := strings.Cut(string(*r), ":"); found {
      s = x
   }
   return
}

func (r *FactKey) Name() (n string) {
   if _, x, found := strings.Cut(string(*r), ":"); found {
      n = x
   } else {
      n = string(*r)
```
```
func TestFactKey(t *testing.T) {
   g := gomega.NewGomegaWithT(t)

   key := FactKey("key")
   g.Expect(key.Source()).To(gomega.Equal(""))
   g.Expect(key.Name()).To(gomega.Equal("key"))

   key = FactKey("test:key")
   g.Expect(key.Source()).To(gomega.Equal("test"))
   g.Expect(key.Name()).To(gomega.Equal("key"))

   key = FactKey("")
   key.With("test", "key")
   g.Expect(key.Source()).To(gomega.Equal("test"))
   g.Expect(key.Name()).To(gomega.Equal("key"))
}
```

addon/application.go
```diff
@@ -223,57 +223,64 @@ func (h *AppFacts) Source(source string) {
...
 //
 // List facts.
-func (h *AppFacts) List() (list []api.Fact, err error) {
-	list = []api.Fact{}
-	path := Path(api.ApplicationFactsRoot).Inject(Params{api.ID: h.appId})
-	err = h.client.Get(path, &list, Param{Key: api.Source, Value: h.source})
+func (h *AppFacts) List() (facts api.FactMap, err error) {
+	facts = api.FactMap{}
+	key := api.FactKey("")
+	key.With(h.source, "")
```
jortel, line 229: Looking the usage here, I wonder if key.Qualify(source) would be more useful then With(source, name)?
Would result in usages like:
```
key := api.FactKey("")
key.Qualify(h.source)
```
And 
```
key := api.FactKey(name)
key.With(h.source)
```

jortel (approved):
Looks great.

PR #400, opened 2023-06-14
:ghost: Hack analysis: Add .java file extension and example markdown incident message for UI use

mturley:
Making incident mock data look a little more real for the UI demo


jortel (approved):
LGTM

PR #401, opened 2023-06-14
:ghost: Hack analysis: Add believable source and target labels for UI demo

mturley:
Followup to #400


jortel (changes requested):
This will result in all of the issue having the same labels.
I would prefer we  create a longer list of targets and using _mod_ id to select them from an array to get a better distribution/sample-set.

mturley:
That's reasonable. Will follow up with that suggested solution if it becomes necessary, for now I'll just add labels manually to unblock the demo.

PR #405, opened 2023-06-16
:bug: Fix filter by businessService name; and id.

jortel:
Filter on  businessService.name not wired up correctly.
Added by ID.

### DIFF
--- a/api/analysis.go
+++ b/api/analysis.go
@@ -690,6 +690,8 @@ func (h AnalysisHandler) Incidents(ctx *gin.Context) {
 // @description - applications
 // @description - application.id
 // @description - application.name
+// @description - businessService.id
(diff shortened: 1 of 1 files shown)


api/analysis.go
```diff
@@ -826,7 +829,8 @@ func (h AnalysisHandler) RuleReports(ctx *gin.Context) {
 // @description - issue.labels
 // @description - application.id
 // @description - application.name
-// @description - businessService.name
+// @description - application.id
```
mansam, line 832: looks like a copy/paste mistake?

jortel: Ah.  Yep.  Fixed.

mansam (approved):
LGTM

PR #407, opened 2023-06-16
:warning: Remove dependence on Jira CreateMeta API

mansam:
On Prem Jira deployments cannot be relied upon to have the CreateMeta endpoint as it can be taxing for large instances and admins may disable it.

Instead of using CreateMeta, we can query project metadata on demand through cheaper endpoints and expose this through the tracker API.

The UI will now need to query /trackers/:id/projects to get a list of projects to display in the select box as the `metadata` field on the Tracker resource will no longer be present. The UI can then pass the project ID through the /trackers/:id/projects/:id2/issuetypes endpoint to get the list of IssueTypes for the project.


jortel (approved):
LGTM

PR #408, opened 2023-06-16
:bug: Fix RuleReport filter on labels.

jortel:
Fix filter on labels.
filter.Resource("issue") returns a filter with ONLY "issue.*" fields. The issueIDs() expects the root filter.

This is leftover of an earlier implementation of reports and just got missed.


jortel:
Included in https://github.com/konveyor/tackle2-hub/pull/409

PR #409, opened 2023-06-17
:bug: Support LIKE filter on lists.

jortel:
Support LIKE filter with _LIST_ (OR) values in the _filter_ package.

```
?filter=category~(This*|That*)

(category LIKE 'This%' OR category LIKE 'That%')
```

Add support for filter on Labels AND using LIKE.  This needs to be done with an INTERSECT so cannot be entirely provided by the _filter_.
```
 ?filter=labels~(This*,That*)
 
SELECT ID FROM Thing WHERE category LIKE 'This%' 
INTERSECT 
SELECT ID FROM Thing WHERE category LIKE 'That%'
```
Adds Field.Expand() greatly simplifies building the query for AND Lists by delegating the building of each _WHERE clause_ to the _filter_ package. Applied pattern everywhere.

Fixes RuleReport filter on labels.

### DIFF
--- a/api/analysis.go
+++ b/api/analysis.go
@@ -1287,7 +1287,7 @@ func (h *AnalysisHandler) appIDs(ctx *gin.Context, f qf.Filter) (q *gorm.DB) {
 			tq := h.DB(ctx)
 			tq = tq.Model(&model.ApplicationTag{})
 			tq = tq.Select("ApplicationID ID")
-			tq = tq.Where(field.SQL())
+			tq = field.Where(tq)
 			q = q.Where("ID IN (?)", tq)
 		}
 	}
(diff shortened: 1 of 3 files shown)


api/analysis.go
```diff
@@ -748,7 +748,7 @@ func (h AnalysisHandler) RuleReports(ctx *gin.Context) {
 	q = q.Joins("Analysis a")
 	q = q.Where("a.ID = i.AnalysisID")
 	q = q.Where("a.ID in (?)", h.analysisIDs(ctx, filter))
-	q = q.Where("i.ID IN (?)", h.issueIDs(ctx, filter.Resource("issue")))
+	q = q.Where("i.ID IN (?)", h.issueIDs(ctx, filter))
```
jortel, line 751: **hint**: Fixes RuleFilter filter on labels.

mansam (approved):
LGTM

PR #410, opened 2023-06-19
:seedling: Add File and RuleSet API tests

aufi:
Moving File and RuleSet-related functions from addon package to binding package and adding basic API tests for both.

Related to https://github.com/konveyor/go-konveyor-tests/issues/2


addon/adapter.go
```diff
@@ -135,14 +135,14 @@ func newAdapter() (adapter *Adapter) {
...
-		},
-		client: client,
+		File:    binding.File{},
+		RuleSet: binding.RuleSet{},
+		client:  client,
 	}
+	//
+	// Setting client in binding package via setter method.
+	adapter.File.SetClient(client)
+	adapter.RuleSet.SetClient(client)
```
aufi, line 145: Since ```client``` is ot available outside the package (binding in this case), I added SetClient methods to allow set the client from addon package. Expecting its deletion when addon is migrated to binding.

jortel: In the binding package, we could just make the `client` in handlers_public_ instead?

aufi: Sure, we can, will update.

jortel (approved):
LGTM

PR #412, opened 2023-06-20
:bug: metrics: count issue export only when successful

mansam:
Fixes https://issues.redhat.com/browse/MTA-809 by moving the counter increment into the specific connector's create implementation, after the ticket has been created successfully.


jortel (approved):
LGTM

PR #413, opened 2023-06-20
:bug: Fix ticket/tracker cascade deletes

mansam:
Cascade delete Tickets when their Application is deleted. Don't allow a Tracker to be removed when it has tickets.

Fixes https://issues.redhat.com/browse/MTA-811


jortel (approved):
LGTM

PR #415, opened 2023-06-20
:bug: Fix labelIDs() query ignoring other issue fields when labels included.

jortel:
The _inner_ query for labels was overwriting `q` which could have contained Issue fields where clauses.

Updated other queries to use `iq` (_inner query_) convention for consistency.


aufi (approved):
LGTM

PR #423, opened 2023-06-27
:seedling: Fix generated model pkg.go.

jortel:
Fix generated model/pkg.go `All()`:
- models need to be in order of dependency (not alphabetical).
- includes unwanted models.

The fix is to build the new All() using the previous migration All().


aufi (approved):
LGTM

PR #427, opened 2023-06-29
:sparkles: Add dependency relationship to RuleSets

mansam:


jortel (dismissed):
LGTM

migration/v6/model/ruleset.go
```diff
@@ -0,0 +1,70 @@
...
+
+	return
+}
+
+//
+// RuleSetDependency represents a ruleset dependency relationship.
+type RuleSetDependency struct {
+	RuleSetID   uint `gorm:"primaryKey"`
+	DependsOnID uint `gorm:"primaryKey"`
+}
```
jortel, line 50: Seems like it would be good to avoid defining the RuleSetDependency model.  curious about why query that table instead of the RuleSet?
Re: BeforeUpdate() - How is Create protected?

jortel (approved):
LGTM

PR #432, opened 2023-07-04
:seedling: add bucket API test

khareyash05:
PR For bucket API test

Work Done
1. CRUD for basic bucket 
2. Test Bucket Directory
3. Test Bucket File


test/api/bucket/api_test.go
```diff
@@ -0,0 +1,84 @@
...
+// 		t.Run("Bucket File Test", func(t *testing.T) {
+// 			err := Client.BucketPut(sample.Path, api.BucketContentRoot)
+// 			if err != nil {
+// 				t.Errorf(err.Error())
+// 			}
+// 		})
+// 	}
+// }
+
+func TestBucketReaper(t *testing.T) {
```
jortel, line 61: This seems outside the scope of the API test.

khareyash05: Hi Jeff Thanks for the review. Actually me and Marek discussed about the Expiration tag in Bucket and wanted to test indirect deletion of the bucket upon Expiration. So we had decided to create a sample bucket and check its expiration

khareyash05: Will do that!

test/api/bucket/api_test.go
```diff
@@ -0,0 +1,205 @@
...
+				t.Errorf("Difference in Path between the buckets %v and %v", gotBucket.Path, expectedBucket.Path)
+			}
+
+			// Delete bucket.
+			assert.Must(t, Bucket.Delete(pathForBucket))
+		})
+	}
+}
+
+func TestBucketFile(t *testing.T) {
```
djzager, line 65: It may turn out to be a rehash of a previous conversation: if we are going to loop through all of the "buckets" and create + delete them, why wouldn't we do all of the operations on created buckets before deleting them. For example, instead of having `TestBucketCRUD` and `TestBucketFile` that both loop through all Samples and have slight differences from there:

1. Create a bunch of buckets, verify each was created correctly
2. List the buckets and make sure that they were all created correctly
3. Do File/Directory related operations on our created buckets
4. Delete the buckets, verify each was deleted correctly
5. Done

I feel like I've had this discussion before and I'm hoping I am not leading you astray. Feel free to point me to any related commentary if it exists.

djzager (commented):
This is a good first effort to tackling the bucket API. Comments for discussion.

test/api/bucket/api_test.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/konveyor/tackle2-hub/test/assert"
+)
+
+func TestBucketCRUD(t *testing.T) {
+	for _, sample := range Samples {
+		t.Run("Bucket CRUD Test", func(t *testing.T) {
+
+			expectedBucket := api.Bucket{
+				Path: sample.Path,
+			}
```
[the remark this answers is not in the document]
khareyash05: no such reason David. Just a convention we had been using in the api tests

test/api/bucket/api_test.go
khareyash05: Thank you for the review and I think we can merge them into one function and that it will reduce the number of LOC and the variables we might be using but this might increase the complexity of the code and also the number of dump variables in active state at that time. Also might decrease readability to the future users. What do you think?

djzager: We may be able to leverage something like https://pkg.go.dev/testing#hdr-Subtests_and_Sub_benchmarks to make it so we can clearly separate these tests (for readability) and allow them to run sequentially to prevent us from thrashing the API with create/delete bucket requests. I'll look through what you have and see if that would help us.

binding/bucket.go
```diff
@@ -34,3 +58,57 @@ func (h *Bucket) Delete(path string) (err error) {
...
+	err = h.Client.getDir(file, path)
+	return
+}
+
+func (h *Bucket) PutDir(file io.Writer, path string) (err error) {
+	err = h.Client.putDir(file, path)
+	return
+}
+
+func (h *Bucket) Compress(src string, buf io.Writer) error {
```
djzager, line 72: I'm struggling to understand how we are testing the API with this method.

djzager (commented):
I think your comment about readability is absolutely showing up in this review. It's tough to read it all in one shot and track down what's being handled by the test client and what's actually under test. I think splitting things up this way will allow us to have clarity without repeating ourselves.

binding/bucket.go
khareyash05: This is just the compress method. Shall I add it to the `api/bucket,go`?

test/api/bucket/sample/template_application_import3.csv
```diff
@@ -0,0 +1,6 @@
+Record Type 1,Application Name,Description,Comments,Business Service,Dependency,Dependency Direction,Binary Group,Binary Artifact,Binary Version,Binary Packaging,Repository Type,Repository URL,Repository Branch,Repository Path,Tag Category 1,Tag 1,Tag Category 2,Tag 2,Tag Category 3,Tag 3,Tag Category 4,Tag 4,Tag Category 5,Tag 5,Tag Category 6,Tag 6,Tag Category 7,Tag 7,Tag Category 8,Tag 8,Tag Category 9,Tag 9,Tag Category 10,Tag 10,Tag Category 11,Tag 11,Tag Category 12,Tag 12,Tag Category 13,Tag 13,Tag Category 14,Tag 14,Tag Category 15,Tag 15,Tag Category 16,Tag 16,Tag Category 17,Tag 17,Tag Category 18,Tag 18,Tag Category 19,Tag 19,Tag Category 20,Tag 20
```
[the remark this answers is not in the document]
khareyash05: No difference David. I didn't had binaries to use thatswhy created this replicas. Shall i replace them with something else?

test/api/bucket/sample/template_application_import.csv
```diff
@@ -0,0 +1,6 @@
+Record Type 1,Application Name,Description,Comments,Business Service,Dependency,Dependency Direction,Binary Group,Binary Artifact,Binary Version,Binary Packaging,Repository Type,Repository URL,Repository Branch,Repository Path,Tag Category 1,Tag 1,Tag Category 2,Tag 2,Tag Category 3,Tag 3,Tag Category 4,Tag 4,Tag Category 5,Tag 5,Tag Category 6,Tag 6,Tag Category 7,Tag 7,Tag Category 8,Tag 8,Tag Category 9,Tag 9,Tag Category 10,Tag 10,Tag Category 11,Tag 11,Tag Category 12,Tag 12,Tag Category 13,Tag 13,Tag Category 14,Tag 14,Tag Category 15,Tag 15,Tag Category 16,Tag 16,Tag Category 17,Tag 17,Tag Category 18,Tag 18,Tag Category 19,Tag 19,Tag Category 20,Tag 20
```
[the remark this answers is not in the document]
khareyash05: Oh Okay will fetch binaries from users

test/api/bucket/api_test.go
djzager: https://github.com/konveyor/tackle2-hub/issues/446

khareyash05:
> Thanks for your contribution, @khareyash05 !
> 
> This PR: #447 documents buckets and files. Your effort reminded me this is long overdue.
> 
> Also, wanted to mention this PR: #445 which is a planned iteration that completes the migration of _binding_ support from the _addon_ adapter package to the relatively new _binding_ package.  As part of this, it adds the bucket binding handler. This will conflict with our PR.  Sorry.  As a result, your PR can focus on the testing.
> 
> A few thoughts about testing, some of which @djzager has already mentioned.
> - The bucket resource CRUD should be tested in its own test.
> - The bucket content CRUD should be tested in its own test.
>    - PUT a file.  I strongly Recommend just a text file, does not need to be a CSV or binary.
>    - GET a file.
>    - DELETE a file.
>    - PUT a directory.  Recommend creating a directory tree in /tmp of just text files.
>    - GET a directory (either root or leaf) within the tree.
>    - GET a file within the tree.
>    - DELETE a file within the tree.
>    - DELETE a directory within the tree.
>  
> The client handles all of the complexities which keeps the tests simple and tests the client.
> 
> Testing the garbage collection (reaping) should be separate tests outside of the API test.

Thank you for the detailed explanation and helping me understand new features being added by the addon adapter. Also thank you for taking out time to give me a detailed workflow of the file and directory tests. 

I will update the PR keeping in mind all the things you mentioned.

Just a question, is there any sample text file which you would want to be used?or can i just a lorem ipsum text file?

khareyash05:
Also @jortel Do I need to test `getDir` and `putDir` as well? because for that I will have to tackle the archive thing and might need some more functions from the `binding` package.

jortel:
> 
> Just a question, is there any sample text file which you would want to be used?or can i just a lorem ipsum text file?
Any file will due.

jortel:
> Also @jortel Do I need to test `getDir` and `putDir` as well? because for that I will have to tackle the archive thing and might have to add more functions to the `binding` package.

The short answer is no.

IMHO, the primary goal is to test the Hub (service-side) API.
And, using the richClient handlers accomplishes 2 things:
1. Greatly simplifies all of the tests.
2. Implicitly tests the  `Client`.

Tests using Bucket.Get() and Bucket.Put() with both source and destination being files and directories will achieve (practically) 100% code coverage (including Client.getDir() and Client.putDir() and everything in between).

aufi:
Related to https://github.com/konveyor/tackle2-operator/issues/220

test/api/bucket/api_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+			// Find for the specific bucket and compare Paths as it is a unique value.
+			for _, gotBucket := range gotBuckets {
+				if gotBucket.ID == bucket.ID {
+					if gotBucket.Path != bucket.Path {
+						t.Errorf("Difference in Path between the buckets %v and %v", gotBucket.Path, bucket.Path)
+					}
+				}
+			}
+			fmt.Println("hello")
```
jortel, line 37: hello?

test/api/bucket/api_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+			// Inject Expected Buckets's ID into the BucketRoot.
+			bucketID := binding.Path(api.BucketRoot).Inject(binding.Params{api.ID: bucket.ID})
+
+			// Get specific bucket.
+			gotBucket := api.Bucket{}
+			err = Client.Get(bucketID, &gotBucket)
```
jortel, line 44: Why using the rest client instead of richClient.Bucket.Get()?

jortel:
Suggest using static files and directory tree for the tests instead of creating them would eliminate a lot of unnecessary code and complexity.  See the _file_ test `file/data`.

khareyash05:
> Suggest using static files and directory tree for the tests instead of creating them would eliminate a lot of unnecessary code and complexity. See the _file_ test `file/data`.

Thanks @jortel WIll do that

test/api/bucket/api_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"os"
+	"testing"
+
+	"github.com/konveyor/tackle2-hub/test/assert"
+)
+
+func TestBucketCRUD(t *testing.T) {
+	for _, bucket := range Buckets {
+
+		t.Run("Create Bucket and Compare", func(t *testing.T) {
```
aufi, line 14: Few notes on this test:
- I'm not sure how much it is needed, but likely it is needed to test various sample files Create/Get/Index, but in case of Bucket.Get it should check not only Path, but also content or content length.
- Nit, but when iterating over samples in subtest, add their name/Path in this case (same for line 46).

test/api/bucket/samples.go
```diff
@@ -0,0 +1,20 @@
+package bucket
+
+import (
+	"github.com/konveyor/tackle2-hub/api"
+)
+
+var Buckets = []api.Bucket{
```
aufi, line 7: Just curious, do we need 4 same sample files or it is just a remain from some refactoring?

aufi (commented):
Thanks for PR, added few comments.

test/api/bucket/samples.go
khareyash05: No just few samples. No such significance. Will update

aufi (approved):
Thanks for update Yash, overall LGTM, merging.

PR #434, opened 2023-07-05
:sparkles: Add links to rules and issues reports.

jortel:
Adds `Links` to both: /rules and /applications/:id/issues reports.
hack script adds markdown issue description and links.


mturley (approved):
LGTM

PR #440, opened 2023-07-12
:bug: Don't return issueTypes with the subtask flag set

mansam:
New issues that belong to issue types with the `subtask` flag set (including the "Sub-task" built in issue type) are required to have a parent issue set. We don't support setting the parent issue, so exporting migration waves as issues of such types will fail. To fix this, avoid returning issue types with the `subtask` flag set from the `/trackers/:id/projects/:pid/issuetypes` endpoint.

Fixes https://issues.redhat.com/browse/MTA-870 by removing the option to export as a Sub-task.


jortel (approved):
LGTM

PR #441, opened 2023-07-13
:sparkles: Better support for addon failed. 

jortel:
Better support for addon failed with reported errors (not returned _error_).


mansam (approved):
LGTM

PR #447, opened 2023-07-19
:ghost: Add bucket,file documentation.

jortel:
Add bucket and files documentation.


docs/bucket.md
```diff
@@ -0,0 +1,58 @@
...
+**GET** will return `thing` as an octet-stream. when thing is a directory, it is packaged as a tarball and
+the `X-Directory=Expand` header set. Bucket (content) may be inspected using a Web Browser. A **GET** on 
+a Bucket route with Accept=text/html will return an index.html.
+
+**PUT** will (upload) store `thing` in the bucket. Intermediate directories are automatically created. When 
+`thing` is a directory, the client must package thing as a tarball and set the `X-Directory=Expand` header.
+
+**DELETE** will delete `thing` (but not intermediate directories).
+
+## Files ##
```
djzager, line 33: Isn't this a totally different resource? I don't see anything in `api/file.go` that would indicate it's related to buckets. I think it deserves it's own file so there is hope those looking for it can find it.

docs/bucket.md
```diff
@@ -0,0 +1,58 @@
...
+
+**GET** will return the file based on the Accept header:
+- Accept=applications/json: The `File` resource.
+- Accept=(other): The file (content) as an octet-stream.
+
+**PUT** Create the `File` resource and store the file (content).
+
+**DELETE** will delete the `File` resource and associated file (content).
+
+### Reaping ###
```
djzager, line 50: I think this should be it's own doc, referenced by both buckets and files.

docs/bucket.md
```diff
@@ -0,0 +1,58 @@
...
+Buckets and Files are garbage collected resources.
+They are considered orphans when no longer referenced by another resource. When an orphan is
+detected, it is assigned an expiration (date/time). The expiration is a kind of _grace period_
+intended to support two-phase assignment to an owner or re-assignment.
+
+### Streaming ###
+
+Both Bucket and File content is uploaded using a multipart form.  Field: `File`=octet-stream.
```
djzager, line 58: I don't know that this blurb tells me anything new. Maybe include it in each of the explanations of "PUT"ing buckets|files?

djzager (changes requested):
This is very helpful indeed. Thank you for putting it together. I have a few thoughts. Additionally, I wonder why we don't pull all this knowledge into the swagger directly and figure out a way to render the swagger for user/developer facing docs.

docs/bucket.md
```diff
@@ -0,0 +1,58 @@
...
+
+
+### Content ###
+The Bucket itself has a _virtual_ content subresource that provides GET/PUT/DELETE operations.  
+For example: a route of `/buckets/1/this/that/thing`  `this/that/thing` is the path to the `thing` subresource (file) 
+within the bucket.
+
+**GET** will return `thing` as an octet-stream. when thing is a directory, it is packaged as a tarball and
+the `X-Directory=Expand` header set. Bucket (content) may be inspected using a Web Browser. A **GET** on 
+a Bucket route with Accept=text/html will return an index.html.
```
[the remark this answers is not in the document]
jortel: The index.html is generated.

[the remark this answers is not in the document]
djzager: So, this is making `https://${hub_ip}/buckets/1/this/that/thing` (where thing is a directory) function the same as when I navigate in my browser to `file:///some/path/on/local/filesystem`? It reads as though you are simply returning an `/this/that/thing/index.html`.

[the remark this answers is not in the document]
djzager: Maybe it would be less confusing to say "return a directory index"?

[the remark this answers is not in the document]
jortel: Yeah, that seems clearer.

docs/bucket.md
jortel: Updated as suggested.

docs/bucket.md
jortel: Split up.

djzager (approved):

aufi (commented):
Nicely written!

PR #450, opened 2023-07-21
:warning: Use the new seeding library to seed rulesets

mansam:
* Migration v7 sets up the new unique UUID columns and deletes the old RuleSets
* For the time being, the seeds are cloned into the Hub image from the tackle2-seed repository.
* Seeds only run if the version number in the seed document matches the DB migration version
* Seeds only run if the seed digest differs from that stored in the `.hub.db.seed` setting.

Seeding operates via the following logic:

* If a model with the same UUID as a seed resource exists, check if it is being renamed. If it is being renamed, ensure that the target name is free by renaming any model that already has that name. Then proceed to update the original model.
* If a model with the same UUID does not exist, look to see if one with the same name already exists.
    * If one does not already exist, create a new model. 
    * If one already exists and it was not created by the Hub (`CreateUser != ""`), rename it to free up the name and then create a new model.
    * If one already exists and it was created by the Hub (`CreateUser == ""`), commandeer it.


jortel (approved):
Looks great, Sam.

PR #451, opened 2023-07-21
:bug: Hide Epic issue type for On Prem trackers

mansam:
The Epic issue type has an additional required field for Server/Datacenter instances. We don't currently have a solution in place for dealing with additional required fields, so hide that issue type on on-prem instances for now.

Fixes https://issues.redhat.com/browse/MTA-873


jortel (approved):
LGTM

PR #454, opened 2023-07-25
Addon binding task report refactored.

jortel:


jortel:
Taking another approach.

PR #457, opened 2023-07-25
:sparkles: Add raw methods for error and activity reporting by addons.

jortel:
Add _raw_ methods used by addons to report both errors and activity. The reason is to reduce the number of API calls when reporting large numbers.
For example, the _addon.command_ package reports stdout/stderr for failed commands which can be hundreds (or thousands of lines).  Calling Activity() hundreds of times result in hundreds of API calls.
Another example is the reported analyzer errors.

Another aspect is to formalize multi-line activity.  This PR codifies the convention of:
```
- This is a list of:
- > One
- > Two
- > Three
```
and, simplifies the API.

The tackle2-addon _command_ package will need to be updated once this merges to use RawActivity() for stdout/stderr. reporting.
The tackle2-addon-analyzer will need to updated to use both methods in select places as well.

### DIFF
--- a/addon/task.go
+++ b/addon/task.go
@@ -6,6 +6,8 @@ import (
 	"github.com/konveyor/tackle2-hub/api"
 	"github.com/konveyor/tackle2-hub/binding"
 	"github.com/konveyor/tackle2-hub/task"
+	"gopkg.in/yaml.v2"
+	"strings"
 )
 
 //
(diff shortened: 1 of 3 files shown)


addon/task.go
```diff
@@ -67,7 +69,6 @@ func (h *Task) Variant() string {
 //
 // Started report addon started.
 func (h *Task) Started() {
-	h.Load()
```
jortel, line 70: Moved Load() to adapter.Run().  Makes more sense.

mansam (approved):

PR #459, opened 2023-07-25
Make UUID field a pointer

mansam:
Fixes an issue where new objects were being created with a UUID of the empty string instead of a null, which was interfering with the uniqueness test.


jortel (approved):
LGTM

PR #460, opened 2023-07-25
:bug: Only create connector once to fix ident mangling

mansam:
Creating the tracker connector inside of the loop results in Decrypt being called on the tracker's Identity once for each ticket to be created. Decrypt is not idempotent, so this scrambles the password on subsequent calls. This was causing batch creation of Jira tickets to fail strangely, as only the first ticket in any batch would be created before the credentials were scrambled.

Fixes https://issues.redhat.com/projects/MTA/issues/MTA-696


jortel (approved):
LGTM

PR #462, opened 2023-07-25
:ghost: SoftError deprecated.

jortel:
The SoftError was originally intended to support addons returning an error that would not be stack traced and not trigger a task (pod) retry.  However, in practice:
- almost all errors should not be retried.
- SoftError has not been consistently adopted.

As a result, SoftError has been all but removed and **deprecated**.  Leaving it in the addon binding for backwards compat.

Further, the adapter Run() will no longer trace caught errors because there cannot be any expectation that addon writers will _wrap_ errors and logging errors in Run() is useless.  The addon returning an error will always result in exit(1) - 
 _application failed_.  The task manager will retry for: code=137 (SIGKILL) Only.

Fixed binding Client.restError() for issue found while testing.

### DIFF
--- a/binding/application.go
+++ b/binding/application.go
@@ -2,6 +2,7 @@ package binding
 
 import (
 	"bytes"
+	"errors"
 	mime "github.com/gin-gonic/gin/binding"
 	liberr "github.com/jortel/go-utils/error"
 	"github.com/konveyor/tackle2-hub/api"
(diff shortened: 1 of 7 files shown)


binding/client.go
```diff
@@ -775,13 +775,16 @@ func (r *Client) restError(response *http.Response) (err error) {
...
 		restError := &NotFound{}
-		err = restError.With(response)
+		restError.With(response)
+		err = restError
 	default:
 		restError := &RestError{}
-		err = restError.With(response)
+		restError.With(response)
+		err = restError
 	}
```
jortel, line 788: **hint**: RestError.With() returned a RestError not the concrete type.  It was _object slicing_.  As a result, restError() was not returned the concrete type as intended.

binding/application.go
```diff
@@ -182,6 +183,16 @@ func (h *AppTags) Add(id uint) (err error) {
 	return
 }
 
+//
+// Ensure ensures tag is associated with the application.
+func (h *AppTags) Ensure(id uint) (err error) {
```
mansam, line 188: :+1:

mansam (approved):
LGTM

PR #464, opened 2023-07-28
:warning: Separate Target and RuleSet models/apis

mansam:


api/target.go
```diff
@@ -0,0 +1,239 @@
...
+
+//
+// TargetHandler handles Target resource routes.
+type TargetHandler struct {
+	BaseHandler
+}
+
+func (h TargetHandler) AddRoutes(e *gin.Engine) {
+	routeGroup := e.Group("/")
+	routeGroup.Use(Required("rulesets"), Transaction)
```
jortel, line 26: targets
Also, needs to be added to the realm.

api/target.go
```diff
@@ -0,0 +1,306 @@
...
+
+//
+// TargetHandler handles Target resource routes.
+type TargetHandler struct {
+	BaseHandler
+}
+
+func (h TargetHandler) AddRoutes(e *gin.Engine) {
+	routeGroup := e.Group("/")
+	routeGroup.Use(Required("targets"), Transaction)
```
jortel, line 26: This is probably practical but expected the Transaction only on:
Create
Update
Delete

migration/v8/model/target.go
```diff
@@ -0,0 +1,20 @@
...
+
+//
+// Target - analysis rule selector.
+type Target struct {
+	Model
+	UUID        *string `gorm:"uniqueIndex"`
+	Name        string  `gorm:"uniqueIndex;not null"`
+	Description string
+	Choice      bool
+	Labels      JSON
```
jortel, line 11: Missing: `gorm:"type:json"`

api/target.go
mansam: I think we've been using it like this everywhere, and relying on the method check in the Transaction handler unless I'm mistaken.

api/ruleset.go
```diff
@@ -268,13 +309,10 @@ func (r *RuleSet) With(m *model.RuleSet) {
 // Model builds a model.
 func (r *RuleSet) Model() (m *model.RuleSet) {
 	m = &model.RuleSet{
-		Kind:        r.Kind,
-		Name:        r.Name,
-		Description: r.Description,
```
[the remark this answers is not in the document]
mansam: Yeah, good idea.

jortel (approved):
Looks great!

api/ruleset.go
```diff
@@ -71,7 +75,16 @@ func (h RuleSetHandler) List(ctx *gin.Context) {
 		h.DB(ctx),
 		clause.Associations,
 		"Rules.File")
-	result := db.Find(&list)
+
+	filter, err := qf.New(ctx,
+		[]qf.Assert{
+			{Field: "labels", Kind: qf.STRING},
```
jortel, line 81: Adding `name` to the filter in the interest of completeness _may_ be a nice to have but can always add later.

api/target.go
jortel: You're correct.  I forgot.

PR #466, opened 2023-08-02
:ghost: Relation renamed.

jortel:
The `Relation` validation was originally added to support:
```
tag=(A,B,C)
```
Where tag = A **and** B **and** C which is implemented as an SQL INTERSECT.
However, it seems the `Labels` use cases are implemented as JSON query.  This PR attempts to provide a more intuitive name.

Replaces:
```
{Field: "Labels", Kind: qf.STRING, Relation: true}, 
```
With
```
{Field: "Labels", Kind: qf.STRING, And: true}, 
```


aufi (approved):
LGTM

PR #468, opened 2023-08-03
:ghost: Shared RuleSet create,update,delete.

jortel:
Refactor RuleSet handler Create(), Update(), Delete() endpoints methods to use the _new_ create(), update(), delete() so that this functionality can be shared with the TargetHandler.

Adds `Target` binding and test.

Also, improves RuleSet _update_ test to include add/delete of rules. This is the most complex part of the updating.

For both Target and RuleSet tests, move file cleanup to deferred function.  I'm not convinced the clean up is necessary but assume the policy here is for tests to have a _zero-footprint_.


aufi (approved):
Looks great!

PR #469, opened 2023-08-06
:bug: Fix dangling ruleset ref to image File.

jortel:
Removes unwanted RuleSet columns that should have been removed in model version:8:
- Custom
- ImageID

The imageID causes problems for the File reaper.  The reaper no longer recognizes RuleSet as a file _owner_ and tries to delete the associated image `File`.  This fails due to the FK constraint.


dymurray (approved):
VISIACK

PR #471, opened 2023-08-08
:seedling: add migrationwave API test

khareyash05:
Added the migration wave API test

### DIFF
--- a/binding/migrationwave.go
+++ b/binding/migrationwave.go
@@ -1,0 +1,50 @@
+package binding
+
+import (
+	"github.com/konveyor/tackle2-hub/api"
+)
+
+//
+// MigrationWave API.
+type MigrationWave struct {
+	client *Client
+}
+
+//
+// Create a MigrationWave.
+func (h *MigrationWave) Create(r *api.MigrationWave) (err error) {
+	err = h.client.Post(api.MigrationWavesRoot, &r)
+	return
+}
+
+//
+// Get a MigrationWave by ID.
+func (h *MigrationWave) Get(id uint) (r *api.MigrationWave, err error) {
+	r = &api.MigrationWave{}
+	path := Path(api.MigrationWaveRoot).Inject(Params{api.ID: id})
+	err = h.client.Get(path, r)
+	return
+}
+
+//
+// List Reviews.
+func (h *MigrationWave) List() (list []api.MigrationWave, err error) {
+	list = []api.MigrationWave{}
+	err = h.client.Get(api.MigrationWavesRoot, &list)
+	return
+}
+
+//
+// Update a MigrationWave.
+func (h *MigrationWave) Update(r *api.MigrationWave) (err error) {
+	path := Path(api.MigrationWaveRoot).Inject(Params{api.ID: r.ID})
+	err = h.client.Put(path, r)
+	return
+}
+
+//
+// Delete a MigrationWave.
+func (h *MigrationWave) Delete(id uint) (err error) {
+	err = h.client.Delete(Path(api.MigrationWaveRoot).Inject(Params{api.ID: id}))
+	return
+}
--- a/binding/migrationwave.go
+++ b/binding/migrationwave.go
@@ -27,7 +27,7 @@ func (h *MigrationWave) Get(id uint) (r *api.MigrationWave, err error) {
 }
 
 //
-// List Reviews.
+// List MigrationWaves.
 func (h *MigrationWave) List() (list []api.MigrationWave, err error) {
 	list = []api.MigrationWave{}
 	err = h.client.Get(api.MigrationWavesRoot, &list)
--- a/test/api/migrationwave/api_test.go
+++ b/test/api/migrationwave/api_test.go
@@ -10,6 +10,35 @@ import (
 
 func TestMigrationWaveCRUD(t *testing.T) {
 	for _, r := range Samples {
+		createdApps := []api.Application{}
+		for _, app := range r.Applications {
+			expectedApp := api.Application{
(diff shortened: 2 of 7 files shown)


binding/migrationwave.go
```diff
@@ -0,0 +1,50 @@
...
+// Get a MigrationWave by ID.
+func (h *MigrationWave) Get(id uint) (r *api.MigrationWave, err error) {
+	r = &api.MigrationWave{}
+	path := Path(api.MigrationWaveRoot).Inject(Params{api.ID: id})
+	err = h.client.Get(path, r)
+	return
+}
+
+//
+// List Reviews.
```
jortel, line 30: List MigrationWaves

test/api/migrationwave/api_test.go
```diff
@@ -0,0 +1,157 @@
...
+	// List MigrationWaves.
+	got, err := MigrationWave.List()
+	if err != nil {
+		t.Errorf(err.Error())
+	}
+
+	// Compare contents of migration waves.
+	for _, createdMigrationWave := range createdMigrationWaves {
+		found := false
+		for _, retrievedReview := range got {
```
jortel, line 122: retrievedWave

test/api/migrationwave/api_test.go
```diff
@@ -0,0 +1,157 @@
...
+			if assert.FlatEqual(createdMigrationWave.ID, retrievedReview.ID) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			t.Errorf("Expected review not found in the list: %v", createdMigrationWave)
```
jortel, line 129: Expected wave

khareyash05:
Oh missed that when using the same elements as Reviews. Will Update all. Thank you for the review.

aufi:
Related to https://github.com/konveyor/operator/issues/220

test/api/migrationwave/api_test.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		for _, stakeholderGroup := range createdMigrationWave.StakeholderGroups {
+			assert.Must(t, StakeholderGroup.Delete(stakeholderGroup.ID))
+		}
+		assert.Must(t, MigrationWave.Delete(createdMigrationWave.ID))
+	}
+}
+
+func AssertEqualMigrationWaves(t *testing.T, got *api.MigrationWave, expected api.MigrationWave) {
```
aufi, line 150: Add checks for associated resources (Applications, Stakeholders, StakeholderGroups) to check all attributes sent via API.

test/api/migrationwave/api_test.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		for i, stakeholderGroup := range r.StakeholderGroups {
+			expectedStakeholderGroup := api.StakeholderGroup{
+				Name:        stakeholderGroup.Name + strconv.Itoa(i),
+				Description: "Sample Stakeholder Group",
+			}
+			assert.Must(t, StakeholderGroup.Create(&expectedStakeholderGroup))
+		}
+		assert.Must(t, MigrationWave.Create(&r))
```
aufi, line 109: Consider order of MigrationWave related objects deletion - try avoid deleting applications associated with an existing migrationwave, instead delete resources in an opposite order than in its creation, so 1. migration wave, then others.

aufi (commented):
Overall looks good to me, added few comments. Thanks!

test/api/migrationwave/api_test.go
khareyash05: Will Keep that in mind.

aufi:
And update API test matrix (as in previous PRs)..

test/api/migrationwave/api_test.go
```diff
@@ -0,0 +1,197 @@
...
+			}
+			assert.Must(t, Application.Create(&expectedApp))
+		}
+
+		for _, stakeholder := range r.Stakeholders {
+			expectedStakeholder := api.Stakeholder{
+				Name:  stakeholder.Name,
+				Email: "sample@example.com",
+			}
+			assert.Must(t, Stakeholder.Create(&expectedStakeholder))
```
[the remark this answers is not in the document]
khareyash05: Will double check on my local env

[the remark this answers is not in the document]
khareyash05: Ya that seems an issue.Noticed how. Thanks for the check will update them

aufi:
Thanks for update, but I'm afraid I'm not OK with the latest change. Plus one on passing test and not leaving orphaned database records :+1: 

But not creating dependencies before creating migration wave doesn't looks right to me. The side-effect that the Create action on Migration wave creates dependencies from its Refs is something I don't see as correct (not saying it is a bug, just an edge case), not sure how my colleagues feel about this, but for my approval, please create (and delete) dependent objects for the migration wave manually. (that should match to typical use-case in UI where user creates migration wave first and then adds applications to the migration wave)

test/api/migrationwave/api_test.go
```diff
@@ -11,29 +10,6 @@ import (
 
 func TestMigrationWaveCRUD(t *testing.T) {
 	for _, r := range Samples {
-		for _, app := range r.Applications {
-			expectedApp := api.Application{
-				Name:        app.Name,
```
aufi: Just add ID here before the Name

aufi (approved):
I have a small problem with creating dependencies in API call together with the MigratioWave, but since API allows it, not a blocker. LGTM.

khareyash05:
Let's discuss this over call. Might have missed something.

PR #473, opened 2023-08-12
:seedling: add tracker API Test

khareyash05:
Added a Tracker API Test

### DIFF
--- a/test/api/tracker/api_test.go
+++ b/test/api/tracker/api_test.go
@@ -1,0 +1,115 @@
+package tracker
+
+import (
+	"testing"
+
+	"github.com/konveyor/tackle2-hub/api"
+	"github.com/konveyor/tackle2-hub/test/assert"
+)
+
+func TestTrackerCRUD(t *testing.T) {
+	for _, r := range Samples {
+		t.Run("Tracker CRUD", func(t *testing.T) {
+			// Create a sample identity for the tracker.
+			identity := api.Identity{
+				Kind: r.Kind,
+				Name: r.Identity.Name,
+			}
+			assert.Must(t, Identity.Create(&identity))
+			// Copy the identity name to the tracker.
+			r.Identity.ID = identity.ID
+
+			// Create a tracker.
+			assert.Must(t, Tracker.Create(&r))
+
+			// Get.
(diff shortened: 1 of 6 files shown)


test/api/tracker/api_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+			projectsList, err := Tracker.ListProjects(r.ID)
+			if err != nil {
+				// check for type of service (maybe Connected?)
+				// if API service then print Not connected to Jira and pass the test
+				// else fail the test
+				if r.Connected == false {
+					t.Logf("Not connected to Jira(Thus passing the API Test)")
+				}
+				if r.Connected == true {
```
aufi, line 81: Thinking why not use ```} else {``` here (or do you want extent the condition with more checks?)

test/api/tracker/api_test.go
```diff
@@ -0,0 +1,192 @@
...
+	if got.URL != expected.URL {
+		t.Errorf("Different Url Got %v, expected %v", got.URL, expected.URL)
+	}
+	if got.Kind != expected.Kind {
+		t.Errorf("Different Kind Got %v, expected %v", got.Kind, expected.Kind)
+	}
+	if got.Connected != expected.Connected {
+		t.Errorf("Different Connected Got %v, expected %v", got.Connected, expected.Connected)
+	}
+	if got.LastUpdated != expected.LastUpdated {
```
aufi, line 183: Just a note on this field, I'd prefer not include LastUpdated field in Equality check (it is similar to e.g. CreatedAt which we don't assert neither).

aufi (commented):
Added few notes, but overall looks good!

aufi:
Related to https://github.com/konveyor/operator/issues/220

test/api/tracker/api_test.go
khareyash05: No I was thinking if this condition would suffice or if we would need more conditions later. Will update.Thanx

test/api/tracker/api_test.go
khareyash05: Will update

aufi (approved):
LGTM

PR #475, opened 2023-08-16
:sparkles: Add support for download static analysis

jortel:


settings/hub.go
```diff
@@ -6,26 +6,27 @@ import (
...
+	EnvTaskSA             = "TASK_SA"
+	EnvTaskRetries        = "TASK_RETRIES"
+	EnvFrequencyTask      = "FREQUENCY_TASK"
+	EnvFrequencyReaper    = "FREQUENCY_REAPER"
+	EnvDevelopment        = "DEVELOPMENT"
+	EnvBucketTTL          = "BUCKET_TTL"
+	EnvFileTTL            = "FILE_TTL"
+	EnvAppName            = "APP_NAME"
+	EnvDisconnected       = "DISCONNECTED"
+	EnvAnalysisReportPath = "ANALYSIS_REPORT_PATH"
```
jortel, line 29: **hint**: just added EnvAnalysisReportPath

test/tar/tar_test.go
```diff
@@ -0,0 +1,65 @@
+package tar
```
aufi, line 1: Thanks for adding unit test tar writer, but the placement of this test migh need to be updated.

Most of unit tests are located directly in the package they test (like auth), so Makefile test target [0] differentiates unit and other tests based on the package.  I'd suggest move this directory content into top-level tar directory (or update Makefile, as you prefer) to ensure the test is executed by make test.

[0] https://github.com/konveyor/tackle2-hub/blob/main/Makefile#L126-L128

aufi (approved):
Added few minor questions, but overall LGTM.

PR #477, opened 2023-08-19
:seedling: add ticket API test

khareyash05:
Add the ticket API Test

### DIFF
--- a/test/api/ticket/api_test.go
+++ b/test/api/ticket/api_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/konveyor/tackle2-hub/api"
+	TrackerSamples "github.com/konveyor/tackle2-hub/test/api/tracker"
 	"github.com/konveyor/tackle2-hub/test/assert"
 )
 
@@ -25,15 +26,7 @@ func TestTicketCRUD(t *testing.T) {
 			assert.Must(t, Identity.Create(&identity))
 
 			// create a sample tracker for the ticket.
-			tracker := api.Tracker{
-				Name: r.Tracker.Name,
-				URL:  "www.sample.com",
-				Kind: "jira-onprem",
-				Identity: api.Ref{
-					ID:   1,
-					Name: "Sample Identity",
-				},
-			}
+			tracker := TrackerSamples.Samples[0]
 			assert.Must(t, Tracker.Create(&tracker))
 
 			// Create a sample ticket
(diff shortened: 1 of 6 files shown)


test/api/ticket/samples.go
```diff
@@ -0,0 +1,22 @@
+package ticket
+
+import (
+	"github.com/konveyor/tackle2-hub/api"
+)
+
+var Samples = []api.Ticket{
+	{
```
aufi, line 8: I'm not sure about typical values of the Ticket, but values of most of fields below might need be updated. Need find some example values.

test/api/ticket/api_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+			// Create a sample identity for the tracker
+			identity := api.Identity{
+				Name: "Sample Identity",
+				Kind: "Sample Kind",
+			}
+			assert.Must(t, Identity.Create(&identity))
+
+			// create a sample tracker for the ticket.
+			tracker := api.Tracker{
```
aufi, line 28: The Tracker fields might be better re-used from from Tracker API test Samples instead of writing it here as well as in List test.

test/api/ticket/api_test.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+		assert.Must(t, Tracker.Create(&tracker))
+
+		// Create a sample ticket
+		assert.Must(t, Ticket.Create(&r))
+		createdTickets = append(createdTickets, r)
+	}
+
+	// List Tickets.
+	got, err := Ticket.List()
```
aufi, line 101: There two additional filters in List actions - by application and tracker - https://github.com/konveyor/tackle2-hub/blob/main/api/ticket.go#L74-L79. Maybe it is not required for simple test, but might be worth for adding those.

test/api/ticket/api_test.go
khareyash05: Shall i run a loop for tracker samples ?

khareyash05: or shall i use only one tracker from the sample?

aufi: One tracker from sample should be enough!

test/api/ticket/api_test.go
```diff
@@ -0,0 +1,164 @@
...
+			createdTrackers := []api.Tracker{}
+			for _, tracker := range TrackerSamples.Samples {
+				// Create a sample identity for the tracker
+				identity := api.Identity{
+					Name: tracker.Identity.Name,
+					Kind: tracker.Kind,
+				}
+				assert.Must(t, Identity.Create(&identity))
+				createdIdentities = append(createdIdentities, identity)
+				assert.Must(t, Tracker.Create(&tracker))
```
aufi, line 31: It looks that the created identity is not assigned to tracker, I'd consider adding something like ```tracker.identitty = identity``` just before creating it.

test/api/ticket/api_test.go
```diff
@@ -0,0 +1,164 @@
...
+				createdTrackers = append(createdTrackers, tracker)
+			}
+
+			// Create a sample ticket
+			assert.Must(t, Ticket.Create(&r))
```
aufi, line 36: The same with quite unreliable link from Ticket to Tracker (ID=1) when one line assignment could set it. But not merge blocker in this case.

test/api/ticket/api_test.go
```diff
@@ -0,0 +1,164 @@
...
+			assert.Must(t, Tracker.Create(&tracker))
+			createdTrackers = append(createdTrackers, tracker)
+		}
+
+		// Create a sample ticket
+		assert.Must(t, Ticket.Create(&r))
+		createdTickets = append(createdTickets, r)
+
+		// List Tickets.
+		got, err := Ticket.List()
```
aufi, line 95: There is no filtering parameter (like Application or Tracker ID), the filter is on API side in List action. Unless filter parameters are sent to API, I'm not sure how the filtering is tested.

aufi (changes requested):
Thanks for update, please check the Ticket List action with filter parameters like Tracker and Application.

aufi (approved):
Thanks for update, LGTM.

PR #482, opened 2023-09-08
:seedling: Update CLI tool login via Hub

aufi:
Login endpoint provided by older Keycloak versions used by CLI tool is not available in current Konveyor version (>=0.2).

Replacing Keycloak get auth token action with Hub's /auth/login endpoint.

Tackle1 export login keeps the original Keycloak login.

### DIFF
--- a/hack/tool/tackle
+++ b/hack/tool/tackle
@@ -82,6 +82,21 @@ def debugPrint(str):
     if args.verbose:
         print(str)
 
+def getHubToken(host, username, password):
+    print("Getting auth token via Hub from %s" % host)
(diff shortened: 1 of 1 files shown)


aufi:
Tested with Konveyor 0.2, example:
```
$ ./tackle -wv export
Loading config from: ./tackle-config.yml
Getting auth token via Hub from https://mta-...
Got access token: eyJhbGciOiJSUzI1NiIsInR5cCIgOiAiSldUIiw...
Exporting Tackle 2 objects (this might take a while..)
...
```

jortel (approved):
LGTM

PR #491, opened 2023-09-21
:warning: Disable seed versioning

mansam:
The seed versioning capability is not necessary and complicates keeping the seed and hub repos in sync. If this is needed in the future, the design should be reevaluated.


aufi (approved):
Don't know all the background, but LGTM.

Issue #428, opened 2023-06-30
Analysis Archiving


jortel:
Goal.

When a new analyses is created, much of the previous report can/should be archived/summarized.
What does the archived analysis need to support:
- When/Who analyzed (history)?
- Effort burn-down?
- Number of issues burn-down? by category?
- Number of incidents burn-down?

Proposed archived information:
- createTime
- createUser
- effort
- []issues*:
  - ruleset
  - rule
  - name
  - description?
  - effort
  - category
  - number of incidents. 

*The storage for this would be relatively small.  The Incidents are the big data stored.  Each analysis will only have 1 issue per rule( that produced an issue).  This will likely be ~40-80 (or less) on average.

---

Proposal

Delete associated rows in Issue, Incident & TechDependency.
Store the _summary_ in Analysis.Summary JSON.

mansam:
This approach sounds good to me.

PR #493, opened 2023-09-26
:sparkles: Analysis archiving

jortel:
Add analysis archiving.

Adds (2) fields to the Analysis:
- Archived (bool)
- Summary JSON which contains the analysis summary when archived.

When an analysis is created, all previously _unarchived_ (archived=FALSE) resources will be archived. There should not be more than one but the logic will do them all just to be sure.  Archived is:
- Summary is created. This is an array of archived issues.
- Archived flag set.
- Issues deleted (cascade delete incidents).
- TechDependencies deleted.

fixes: #428


migration/v10/model/analysis.go
```diff
@@ -0,0 +1,77 @@
+package model
+
+//
+// Analysis report.
+type Analysis struct {
+	Model
+	Effort        int
+	Archived      JSON             `gorm:"type:json"`
```
jortel, line 8: Added ^

mansam (approved):
LGTM

Issue #426, opened 2023-06-29
:ghost: Migration-models


jortel:
The current approach is to separate models into individual files. Then, only copy the file for the model being updated to the next migration. All other types would be aliased. In reality all related models must also be _copied forward_.  As a result, all of the separate flies has turned out to be less practical.  Further, there is a danger that if all of the related models are not copied into the new migration, a mismatch can occur.  It has been determined that it would be safer for each migration to a complete re-definition of all of the models.

Let's start with v6.
Then, let's refactor v2-v5 for consistency.

To facilitate this, the models should be consolidated back into fewer (5) files.
- core.go
   - Model
   - Setting
   - Bucket
   - BucketOwner
   - File
   - Task
   - TaskGroup
   - TaskReport
   - Identity
   - Proxy
- application.go
   - Application
   - BusinessService
   - Dependency
   - Fact
   - JobFunction
   - MigrationWave
   - Stakeholder
   - StakeholderGroup
   - Tag
   - TagCategory
   - Archetype
   - Ticket
   - Tracker
    - Import
   - ImportTag
  - ImportSummary
- assessment.go
  - Assessment
  - Questionnaire
  - Review
- analysis.go
   - TechDependency
   - Incident
   - Analysis
   - Issue
   - RuleSet
   - Rule
   - Target

The hack/next-migration needs to be updated to copy all of the models files forward.

PR #497, opened 2023-09-28
:ghost: refactor migrations

jortel:
Process:
1. For each version(3-10): 
   - Copy files from previous version unless redefined (already exists).
   - Remove the type alias (to previous version) in pkg.go.
2. For each version(2-10): 
   - Consolidate files as defined in #426 .
   - Edit files to fix imports  (just 3-4).

fixes: #426


mansam (approved):
LGTM. Glad to be rid of the aliases.

PR #501, opened 2023-10-03
:sparkles: Allow creating assessments "as-is"

mansam:
If the "sections" field of the assessment resource is populated when POSTing an assessment, then the
sections will not be overridden by values from the questionnaire and the assessment will not be autofilled.


jortel (approved):
LGTM

PR #502, opened 2023-10-03
Surface assessment tags on Archetype resource

mansam:
* added `assessmentTags` field to Archetype resource to display the tags that would be inherited by member applications.
* Fixed a bug where an application would be denied membership in an archetype that wasn't a strict subset of another archetype.


aufi (approved):
LGTM

jortel (approved):
LGTM

PR #506, opened 2023-10-05
:bug: Only marshal assessment sections if not empty

mansam:
Marshalling the Assessment resource's `Sections` field unconditionally results in populating the model's `Sections` with the byte string `null`, which breaks the test for empty sections, making it always look like the assessment is being imported with sections "as-is". To fix this, only marshal the sections in the Assessment's `Model()` method if they are not empty. Additionally, replace the `if m.Sections == nil` test with a test for length, which works on nil as well as empty slices.


jortel (approved):
LGTM

PR #508, opened 2023-10-06
:bug: PrepareForArchetype should use criteria and tags

mansam:
Previously PrepareForArchetype was using only the membership criteria to autofill assessments, but the enhancement specifies that it should use the archetype tags as well as the criteria tags.

Fixes https://issues.redhat.com/browse/MTA-1396


jortel (approved):
LGTM

PR #512, opened 2023-10-10
:bug: Protect builtin questionnaires from update/delete

mansam:
Only changing the `required` field is permitted for builtin questionnaires, all other changes and deletion are disallowed.


ibolton336:
Screenshot 2023-10-10 at 11 46 23 AM

Tested and looks great.

ibolton336 (approved):

PR #514, opened 2023-10-11
:seedling: Fix Identity API test Maven sample

aufi:
Updating Maven Identity sample to use (currently) correct "maven" kind instead of "mvn" to make it working with addon https://github.com/konveyor/tackle2-addon-analyzer/blob/61027b29f36ffa5b884ceac81c61d74341bdd762/cmd/mode.go#L151-L153


jortel (approved):
LGTM

PR #520, opened 2023-10-16
:book: Update OpenAPI spec w/ Assessment changes

mansam:
`ArchivedIssue` in the API package being an alias to another type confuses the swag tool. To mitigate this, I had to overwrite the swagger type of that field on the Analysis resource with `object`.  This means that the api documentation for that field will be unhelpfully vague, but it allows the documentation to build.


jortel (approved):
LGTM

aufi (approved):
LGTM

PR #525, opened 2023-10-17
:bug: Bucket getdir

jortel:
The refactor to using the new `tar` package accidentally removed setting the X-Directory:Expand.
As a result, the client is not expanding downloaded directory (tarball).

---

Fixed API tests.  The bucket directory test really needs 2+ files to be accurate. It missed this issue.
Added the task bucket (subresource) test for symmetry.

### DIFF
--- a/api/bucket.go
+++ b/api/bucket.go
@@ -325,11 +325,10 @@ func (h *BucketOwner) getDir(ctx *gin.Context, input string, filter tar.Filter) 
 	if err != nil {
 		_ = ctx.Error(err)
 		return
-	} else {
-		h.Attachment(ctx, pathlib.Base(input)+".tar.gz")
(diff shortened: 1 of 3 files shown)


mansam (approved):

api/bucket.go
```diff
@@ -325,10 +325,10 @@ func (h *BucketOwner) getDir(ctx *gin.Context, input string, filter tar.Filter)
 	if err != nil {
 		_ = ctx.Error(err)
 		return
-	} else {
-		h.Attachment(ctx, pathlib.Base(input)+".tar.gz")
-		ctx.Status(http.StatusOK)
 	}
+	h.Attachment(ctx, pathlib.Base(input)+".tar.gz")
+	ctx.Writer.Header().Set(Directory, DirectoryExpand)
```
jortel, line 330: **hint**: Add header

PR #528, opened 2023-10-17
:bug: Prevent apps from assessed by default on initial creation

ibolton336:
- Prevents the app to appear "assessed" on initial creation. Adds consideration for the case when no assessments are present, leading to an incorrect true value.

### DIFF
--- a/assessment/application.go
+++ b/assessment/application.go
@@ -99,16 +99,19 @@ func (r *ApplicationResolver) Assessed() (assessed bool, err error) {
 		assessed = r.questionnaireResolver.Assessed(r.application.Assessments)
 		return
 	}
+	assessed = false
+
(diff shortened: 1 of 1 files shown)


mansam (changes requested):
This doesn't work. As written, the value of `assessed` starts `false` and the conditional keeps it `false`, and there's no case where it becomes true for an application with assessments.

The bug in the original implementation is that if the application has no assessments and belongs to no archetypes, it falls through to `assessed = true`. A check for a non-zero number of archetypes would solve it. Something like this:

```go
	assessedCount := 0
	for _, a := range archetypes {
		if r.questionnaireResolver.Assessed(a.Assessments) {
			assessedCount++
		}
	}
	assessed = assessedCount > 0 && assessedCount == len(archetypes)
```

PR #529, opened 2023-10-18
:seedling: Delete identity associated with tracker

mguetta1:
Fixing resources cleanup at the end of the test, by sending the identity's ID to `Identity.Delete`  instead of the tracker's ID


jortel:
This cannot be reviewed without at least _something_ in the description.

aufi (approved):
Thanks for the fix!

PR #530, opened 2023-10-18
Add `confidence` and `risk` to Application and Archetype resources.

mansam:
* Add `confidence` and `risk` fields to Application and Archetype resources. Overall confidence calculated using the algorithm from Pathfinder for each assessment, then summed and divided by total number of assessments. Overall risk level is `green` if all assessments are `green`, otherwise the overall risk is `red`, `unknown`, or `yellow` in that priority order if any assessments have that risk level.
* Moved some assessment status logic out of the resource and into the assessment package where it belongs.
* Added ArchetypeResolver to better encapsulate the related behaviors and for symmetry with ApplicationResolver.
* Simplified the Application and Archetype endpoints by replacing several related resource `With*` methods with `WithResolver`.


api/application.go
```diff
@@ -1112,13 +1069,15 @@ type Application struct {
 	Comments        string      `json:"comments"`
 	Identities      []Ref       `json:"identities"`
 	Tags            []TagRef    `json:"tags"`
-	BusinessService *Ref        `json:"businessService"`
+	BusinessService *Ref        `json:"businessService" yaml:"businessService"`
```
mansam, line 1072: Did these while I was here, still need to do a comprehensive sweep in another PR.

jortel (approved):
LGTM

aufi:
@mansam @jortel Shouldn't this change be backported to release-0.3?

Issue #505, opened 2023-10-05
Seeding Targets causes custom targets to be dropped from the UI display order


mansam:
The portion of the seeding mechanism that adds seeded targets to the display order setting is too simplistic, and it fails to include any targets that exist but were not in the seed file. This causes custom targets to disappear from the UI. This needs to be rewritten to merge newly added targets into the existing setting so that custom targets are not dropped.

PR #536, opened 2023-10-19
:bug: Add missing Targets to ui.target.order after seeding

mansam:
This also adds some convenience methods to the `Setting` model to make reading and updating their values somewhat less tedious.

Fixes https://github.com/konveyor/tackle2-hub/issues/505

### DIFF
--- a/seed/target.go
+++ b/seed/target.go
@@ -10,6 +10,8 @@ import (
 	"gorm.io/gorm"
 )
 
+const UITargetOrder = "ui.target.order"
+
 //
 // Target applies Target seeds.
 type Target struct {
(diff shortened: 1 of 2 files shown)


seed/target.go
```diff
@@ -92,17 +93,51 @@ func (r *Target) Apply(db *gorm.DB) (err error) {
...
+	}
+	s := model.Setting{}
+	result = db.First(&s, "key", UITargetOrder)
+	if result.Error != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+	ordering := []uint{}
+	_ = s.As(&ordering)
+	known := make(map[uint]bool)
```
jortel, line 123: What if `map[uint]int`  where value is the index (order).

jortel (approved):
LGTM
The container/list seems handy.

aufi (approved):
LGTM

PR #537, opened 2023-10-19
:sparkles: Surface latest analysis effort on Application resource

mansam:


api/application.go
```diff
@@ -1125,6 +1125,13 @@ func (r *Application) With(m *model.Application, tags []model.ApplicationTag) {
 		ref.With(a.ID, "")
 		r.Assessments = append(r.Assessments, ref)
 	}
+
+	if len(m.Analyses) > 0 {
+		sort.Slice(m.Analyses, func(i, j int) bool {
+			return m.Analyses[i].CreateTime.Before(m.Analyses[j].CreateTime)
+		})
+		r.Effort = m.Analyses[len(m.Analyses)-1].Effort
+	}
```
jortel, line 1134: Consider sorting by ID to be consistent with other queries (including the First(), Last() method on db).

jortel (approved):
Including (not omitting) the Analysis relation then sorting them is a cleaver idea :-).
I also think adding these columns in the resource makes sense for now.  We can revisit the "report/" concept later.

PR #543, opened 2023-10-26
:bug: Add `yaml:"inline"` to Resource field tags

mansam:
The yaml encoder/decoder needs to be explicitly told inline the `Resource` struct's fields on each of the API resources, otherwise attempts to bind requests that include Resource fields (`id`, `createTime`, `createUser`, `updateUser`) fail.


jortel (approved):
LGTM

PR #549, opened 2023-11-02
:bug: Fix yaml task data

jortel:
Fix JSON encoding of resources fields of type _any_.
When the `content-type` is YAML, the decoded resource fields of type _any_ will be unmarshalled as `map[any]any`.  This is legal for YAML but not JSON.  As a result, the JSON encoding fails and the field value is NOT stored.


aufi (approved):
LGTM

mansam (approved):
LGTM

PR #550, opened 2023-11-07
:sparkles: add addon Adapter.Log.

jortel:
Add support for Adapter.Log to complement addon.Activity.  This provides convenient way for addons to include information in the pod log but not the task activity (which is also logged).  Add writers do not need to construct their own logger.

Used in addons as:
```
addon.Log.Info("Something secret.")
```


mansam (approved):
LGTM

Issue #538, opened 2023-10-20
CSV template for import doesn't contain columns for Owner and Contributors


rromannissen:
The CSV template used to import applications at scale doesn't contain columns to add information about Owner and Contributors for a given application. Having to add that information manually on a per application basis doesn't scale when managing hundreds or thousands of applications. 

The proposal would be to add the following columns:

- Owner: Single value
  - Format: Name <valid_email>
  - Example: Hari Seldon \<hseldon@foundation.com\>
- Contributors: Comma separated list
  - Format: Name1 <valid_email1>, Name2 <valid_email2>
  - Example: Hober Mallow \<hmallow@foundation.com\>, Golan Trevize \<gtrevize@foundation.com\>

rromannissen:
Please bear in mind that the CSV template  available under the *Manage Imports* option at the Application Inventory will have to be updated accordingly once this change is implemente.

dymurray:
Closing due to inactivity or age, please reopen if you feel this was done in error

PR #551, opened 2023-11-07
:warning: Add "owner" and "contributors" columns to application import CSV format.

mansam:
Adds two fields to the import CSV immediately before tags. One is for the Owner and should be empty or contain a single value in the form 'John Doe <jdoe@example.com>'. The second is for contributors, and should be empty or a list of one or more values in the previous format, separated by commas. As this would be a comma-delimited field within a CSV, the whole value should be surrounded by quotes. This breaks compatibility with any existing import CSVs since it adds columns in the middle (necessary since the number of tags is unbounded and as such the tag list must come last).

Fixes https://github.com/konveyor/tackle2-hub/issues/538

@jortel Do you think we can get away with adding this to migration 11, or do you think I should create a new one?


aufi (approved):
LGTM, just tests need to be updated - https://github.com/konveyor/tackle2-hub/tree/main/test/api/importcsv

aufi:
@mansam I think Stakeholders created by CSV import test might need to be cleaned-up after the CSV import test (to not fail other tests).

fabianvf:
Why did this merge with failing CI?

For future reference, we should merge the API test updates and the PR that breaks CI simultaneously. This change broke the integration CI across all repos. We can make changes to the API tests repo (https://github.com/and ensure it works in this branch by adding

            Go tests PR: <PR number>

to a PR description, which will make it run against the proposed changes to ensure the CI stays green across the org as we merge these changes.

Issue #557, opened 2023-11-14
Update report dependencies endpoint to group by name only


sjd78:
For UI changes https://github.com/konveyor/tackle2-ui/issues/1338 and https://github.com/konveyor/tackle2-ui/issues/1340, the dependency table needs to display dependencies grouped by name only.  It currently displays rows based on name and sha.

Endpoint `/analyses/report/dependencies` is what is currently being used.  It returns struct:
```golang
//
// DepReport REST resource.
type DepReport struct {
	Provider     string   `json:"provider"`
	Name         string   `json:"name"`
	Version      string   `json:"version"`
	SHA          string   `json:"sha"`
	Labels       []string `json:"labels"`
	Applications int      `json:"applications"`
}
```
Please adjust the response to group dependencies by name only.  The `SHA` and `Version` fields are no longer needed.   The `Applications` count and `Labels` summary fields should also be adjusted.

Endpoint `/analyses/reports/dependencies/applications` is used by the application table on a dependency's details view.  This endpoint works as needed, but the response object has the dependency's sha named "rule":
```golang
//
// DepAppReport REST resource.
type DepAppReport struct {
	ID              uint   `json:"id"`
	Name            string `json:"name"`
	Description     string `json:"description"`
	BusinessService string `json:"businessService"`
	Dependency      struct {
		ID       uint     `json:"id"`
		Provider string   `json:"provider"`
		Name     string   `json:"name"`
		Version  string   `json:"version"`
		SHA      string   `json:"rule"`
		Indirect bool     `json:"indirect"`
		Labels   []string `json:"labels"`
	} `json:"dependency"`
}
```
Please rename the SHA field's json name to "sha".

sjd78:
cc @ibolton336

jortel:
Is the filter on:
- version
- sha

still needed?

sjd78:
> Is the filter on:
> 
>     * version
> 
>     * sha
> 
> 
> still needed?

I don't think so... At the very least, they will not have a use on the dependencies table.

PR #558, opened 2023-11-14
:bug: DepReport aggregated by name.

jortel:
Update /analyses/report/dependencies to group by (provider, name) instead of (name, version, sha). Currently, provider is always "".

Removes DepReport fields:
- version
- sha

Also fixes DepAppReport.SHA JSON field name.

fixes: #557


sjd78:
Successfully tested with UI on https://github.com/konveyor/tackle2-ui/pull/1536.

PR #559, opened 2023-11-15
:seedling: Add api_hub_tests_ref to CI on PRs

aufi:
In order to run E2E API tests with tackle2-hub API tests on matching branches, adding name of PR branch as a variable passed to CI workflow.

Depends on https://github.com/konveyor/ci/pull/23

Related to https://github.com/konveyor/ci/issues/22 and https://github.com/konveyor/go-konveyor-tests/pull/61

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -94,3 +94,4 @@ jobs:
     uses: konveyor/ci/.github/workflows/global-ci.yml@main
     with:
       component_name: tackle2-hub
+      api_hub_tests_ref: ${{ github.head_ref }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -94,4 +94,4 @@ jobs:
     uses: konveyor/ci/.github/workflows/global-ci.yml@main
     with:
       component_name: tackle2-hub
-      api_hub_tests_ref: ${{ github.head_ref }}
+      api_hub_tests_ref: ${{ github.ref_name }}


.github/workflows/main.yml
```diff
@@ -94,3 +94,4 @@ jobs:
     uses: konveyor/ci/.github/workflows/global-ci.yml@main
     with:
       component_name: tackle2-hub
+      api_hub_tests_ref: ${{ github.head_ref }}
```
fabianvf, line 97: I think we might want `github.ref_name` here instead, I think `head_ref` will reference just the PR branch while `ref_name` will reference the merged PR

aufi: This workflow should affect only PRs. Once the change is merged the branch name (main and e.g. release-0.3) should be driven by already existing API_TESTS_REF.

fabianvf: I think the difference is `head_ref` targets `/refs/pull/<PR #>/head` whereas `ref_name` targets `/refs/pull/<PR #>/merge`. Generally we want to test with merge so that we get the results of the tests _if that PR merged_. Can help catch issues that are introduced on the main branch after a PR is introduced. Let me know if I'm missing something or talking past you

aufi: Ah, good point, I misuderstood it before, thanks! I will update.

fabianvf (approved):
LGTM

PR #564, opened 2023-11-27
Drop Tackle 1.2 export from CLI

aufi:
Removing export Tackle 1.2 function from CLI.

Use older CLI versions if the Tackle 1.2 export is still needed (branch `release-0.2` or older).


aufi:
This should be quite safe change.

PR #566, opened 2023-11-30
:bug: [MTA-1455] Fix typo in tackle-migrator role yaml

mansam:
The migrator role was not able to access the `targets` resource correctly due to a keyname typo in `roles.yaml`. The incorrectly named `verb` key was ignored by the unmarshaller, which resulted in there being no verbs permitted for the resource. This fixes the typo in the yaml, turns on strict unmarshalling in the `LoadUsers` and `LoadRoles` functions, and then adds a unit test to catch future typos.

The tests will catch key typos, typos in http verbs, and typos in user role mappings.

Fixes https://issues.redhat.com/browse/MTA-1455

### DIFF
--- a/auth/role.go
+++ b/auth/role.go
@@ -1,11 +1,12 @@
 package auth
 
 import (
+	"io"
+	"os"
+
 	liberr "github.com/jortel/go-utils/error"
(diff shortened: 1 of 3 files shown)


dymurray (approved):
LGTM

Issue #572, opened 2023-12-05
Tolerate duplicate analysis dependencies. 


jortel:
Reported analysis dependencies my contain duplicates. This needs to be tolerated.
Some use cases:
- The same dependency reported on an analysis as both direct and indirect.
- The same dependency reported as indirect by multiple dependencies (transient).

PR #573, opened 2023-12-05
:bug: TechDependencies deduplicated with direct taking priority.

jortel:
TechDependencies deduplicated with direct taking priority.
The solution is to sort the dependencies with DIRECT first. Then relax the Create() to ignore dups on conflict.
The dependencies are not a memory concern.

fixes: #572


mansam (approved):
LGTM

PR #575, opened 2023-12-06
:bug: Questionnaire validation should tolerate `order` fields having value 0

mansam:
The `required` binding validation can't differentiate an int with value 0 from a value that isn't present. Update `order` fields on questionnaire structs to be represented as a pointer that must not be nil. Also disable `excluded_with` validations that don't work on 0 length slices.


djzager (approved):
VISACK

mansam:
Test failures are known to be unrelated to this PR. Merging.

PR #576, opened 2023-12-12
:seedling: If seed dir exists on the filesystem, don't clone it in the dockerfile

fabianvf:
This was the simplest way to do it I think, without requiring Makefile updates to pass along build-args and dealing with various values. If you just drop tackle2-seed into the project it won't clone it anymore.


fabianvf:
The failing test is from ruleset changes that have broken CI, unrelated to this PR (this PR is the first step to gating that repo so we stop getting breakages)

shawn-hurley (approved):
I am a huge +1 I almost feel like we need to merge this red

jortel (approved):
LGTM

PR #581, opened 2023-12-18
:bug: [MTA-1879] Fix Risk/Confidence calculations to only count responses to required questionnaires.

mansam:
This PR fixes two problems:

1) Application `Risk` and `Confidence` will be `"unknown"` and `0` respectively until all required questionnaires are answered.
2) Answers to archived questionnaires will be ignored while performing the calculation.

Fixes https://issues.redhat.com/browse/MTA-1879

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -2,14 +2,15 @@ package api
 
 import (
 	"encoding/json"
+	"net/http"
+	"sort"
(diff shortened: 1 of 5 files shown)


jortel (approved):
LGTM

PR #585, opened 2023-12-19
:bug: Fix ptr-to-loopvar bug in assessment tags module

mansam:
Fixes an issue which is causing assessment tag autoapply and question include/exclude to fail.


jortel (approved):
LGTM

PR #588, opened 2024-01-05
:ghost: Add questionnare required field to assessment

ibolton336:
This is needed to aid UI in computing status for an assessment given a questionnaire is marked as not required. If an assessment exists for an archived questionnaire on an app or archetype, the assessment status is no longer valid & the assessment will need to be filtered from status consideration. 

https://issues.redhat.com/browse/MTA-1956

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -959,7 +959,7 @@ func (h ApplicationHandler) StakeholdersUpdate(ctx *gin.Context) {
 func (h ApplicationHandler) AssessmentList(ctx *gin.Context) {
 	m := &model.Application{}
 	id := h.pk(ctx)
-	db := h.preLoad(h.DB(ctx), clause.Associations, "Assessments.Stakeholders", "Assessments.StakeholderGroups")
+	db := h.preLoad(h.DB(ctx), clause.Associations, "Assessments.Stakeholders", "Assessments.StakeholderGroups", "Assessments.Questionnaire")
 	db = db.Omit("Analyses")
 	result := db.First(m, id)
 	if result.Error != nil {
--- a/api/archetype.go
+++ b/api/archetype.go
@@ -254,7 +254,7 @@ func (h ArchetypeHandler) Update(ctx *gin.Context) {
 func (h ArchetypeHandler) AssessmentList(ctx *gin.Context) {
 	m := &model.Archetype{}
 	id := h.pk(ctx)
-	db := h.preLoad(h.DB(ctx), clause.Associations, "Assessments.Stakeholders", "Assessments.StakeholderGroups")
+	db := h.preLoad(h.DB(ctx), clause.Associations, "Assessments.Stakeholders", "Assessments.StakeholderGroups", "Assessments.Questionnaire")
 	result := db.First(m, id)
 	if result.Error != nil {
 		_ = ctx.Error(result.Error)
--- a/api/assessment.go
+++ b/api/assessment.go
@@ -164,6 +164,7 @@ type Assessment struct {
 	Status       string                  `json:"status"`
 	Thresholds   assessment.Thresholds   `json:"thresholds"`
 	RiskMessages assessment.RiskMessages `json:"riskMessages" yaml:"riskMessages"`
+	Required     bool                    `json:"required"`
 }
 
 //
@@ -187,6 +188,7 @@ func (r *Assessment) With(m *model.Assessment) {
 	}
 	a := assessment.Assessment{}
 	a.With(m)
+	r.Required = a.Questionnaire.Required
 	r.Risk = a.Risk()
 	r.Confidence = a.Confidence()
 	r.RiskMessages = a.RiskMessages


api/application.go
```diff
@@ -993,7 +993,13 @@ func (h ApplicationHandler) AssessmentList(ctx *gin.Context) {
 	resources := []Assessment{}
 	for i := range assessments {
 		r := Assessment{}
-		r.With(&assessments[i])
```
mansam, line 996: There's already a questionnaire resolver instantiated in this method that you can reuse (line 970), you don't need to instantiate a new one. These assessment-related resolver objects cache the results of their operations, so it's best to reuse them when possible and especially avoid instantiating them in a loop unless its necessary.

api/application.go
```diff
@@ -1061,8 +1067,13 @@ func (h ApplicationHandler) AssessmentCreate(ctx *gin.Context) {
 	if newAssessment {
 		metrics.AssessmentsInitiated.Inc()
 	}
+	questionnaires, err := assessment.NewQuestionnaireResolver(h.DB(ctx))
```
mansam, line 1070: The questionnaire is already loaded from the database on lines 1038-1040, so you can grab the property from that instead of setting up a resolver.

api/application.go
mansam: Also, if you were to add `"Assessments.Questionnaire"` to the list of preloads on line 962, you should be able to grab the `required` property directly from each assessment's linked questionnaire and skip using the resolver here entirely.

api/archetype.go
```diff
@@ -264,7 +264,13 @@ func (h ArchetypeHandler) AssessmentList(ctx *gin.Context) {
 	resources := []Assessment{}
 	for i := range m.Assessments {
 		r := Assessment{}
-		r.With(&m.Assessments[i])
+		questionnaires, err := assessment.NewQuestionnaireResolver(h.DB(ctx))
```
mansam, line 267: See above comments on `ApplicationHandler.AssessmentList`

api/archetype.go
```diff
@@ -330,8 +336,13 @@ func (h ArchetypeHandler) AssessmentCreate(ctx *gin.Context) {
 	if newAssessment {
 		metrics.AssessmentsInitiated.Inc()
 	}
+	questionnaires, err := assessment.NewQuestionnaireResolver(h.DB(ctx))
```
mansam, line 339: Same goes for here as with `ApplicationHandler.AssessmentCreate`

api/assessment.go
```diff
@@ -52,8 +52,15 @@ func (h AssessmentHandler) Get(ctx *gin.Context) {
 		_ = ctx.Error(result.Error)
 		return
 	}
+
```
mansam, line 55: The preload on line 49 should have the Questionnaire relationships on the assessment populated, and you should be able to grab the required property from there rather than setting up a resolver.

api/assessment.go
```diff
@@ -73,10 +80,16 @@ func (h AssessmentHandler) List(ctx *gin.Context) {
 		_ = ctx.Error(result.Error)
 		return
 	}
+	questionnaires, err := assessment.NewQuestionnaireResolver(h.DB(ctx))
```
mansam, line 83: The preload on line 77 should have the Questionnaire relationships on each assessment populated, and you should be able to grab the required property from there rather than setting up a resolver.

api/assessment.go
```diff
@@ -164,11 +177,12 @@ type Assessment struct {
 	Status       string                  `json:"status"`
 	Thresholds   assessment.Thresholds   `json:"thresholds"`
 	RiskMessages assessment.RiskMessages `json:"riskMessages" yaml:"riskMessages"`
+	Required     bool                    `json:"required"`
 }
 
 //
 // With updates the resource with the model.
-func (r *Assessment) With(m *model.Assessment) {
+func (r *Assessment) With(m *model.Assessment, resolver *assessment.QuestionnaireResolver) {
```
mansam, line 185: We've got a convention in the Hub API of creating a method called `WithResolver` to pass the resolver into so that With can continue to just take the model rather than taking an optional parameter. But I think using the resolver here may be redundant anyway based on previous comments.

mansam (changes requested):
Good work! I've got some comments that might let you simplify your approach.

api/application.go
```diff
@@ -994,6 +994,12 @@ func (h ApplicationHandler) AssessmentList(ctx *gin.Context) {
 	for i := range assessments {
 		r := Assessment{}
 		r.With(&assessments[i])
+		if assessments[i].Questionnaire.ID != 0 {
```
ibolton336, line 997: @mansam Wasn't sure if this was the best way here- When trying 

```
Cannot convert 'nil' to type 'Questionnaire'
``` 
I was seeing this type error. Is using the ID here acceptable?

mansam: You're getting that error about converting `nil` because `Assessment.Questionnaire` is a struct instead of a struct pointer. Because it's a struct and not a struct pointer, when the relationship isn't preloaded the struct will be present but all the fields will be zero-valued. So checking if the Questionnaire struct's ID field is 0 like that is a reasonable way to check if it's been preloaded. (Probably that field should be changed to a struct pointer like the rest of the model relationships so that it's more obvious when it hasn't been loaded.)

There are two improvements we can make here. One is that instead of doing this check in the handler logic in all the various places where we need to do it, we can move the check and the assignment into the `Assessment.With` method.

The second is that the zero-value of a boolean is `false`, so we can take advantage of that to do away with the check. `assessment.Required` is going to default to `false` anyway if you don't assign to it because the questionnaire isn't loaded, and the unloaded questionnaire's `Required` will default to false, so the assignment only matters if the Questionnaire is loaded and and required. So these two expressions are equivalent:

```go
if assessment.Questionnaire.ID != 0 {
    assessment.Required = assessment.Questionnaire.Required
}
```
```go
assessment.Required = assessment.Questionnaire.Required
```

(If we later fix assessement.Questionnaire to be a struct pointer, then we'll need to add a check for nil, but for now this works well.)

api/assessment.go
```diff
@@ -159,11 +160,12 @@ type Assessment struct {
...
-	Confidence   int                     `json:"confidence"`
-	Status       string                  `json:"status"`
-	Thresholds   assessment.Thresholds   `json:"thresholds"`
-	RiskMessages assessment.RiskMessages `json:"riskMessages" yaml:"riskMessages"`
+	Risk                  string                  `json:"risk"`
+	Confidence            int                     `json:"confidence"`
+	Status                string                  `json:"status"`
+	Thresholds            assessment.Thresholds   `json:"thresholds"`
+	RiskMessages          assessment.RiskMessages `json:"riskMessages" yaml:"riskMessages"`
+	QuestionnaireRequired bool                    `json:"questionnaireRequired"`
```
mansam, line 168: When you add a field with a camelCased name, be sure to add both a `json` and a `yaml` field tag with the camelCased spelling. (see the `RiskMessages` field  on the line above). This prevents a problem where the yaml encoder will emit it from the api as `camelcased` but the strict decoder will not recognize it because it expects `camelCased`.

nitpick: Lets just call this `Required` for simplicity and for symmetry with the other fields that are ultimately inherited from the questionnaire.

mansam (changes requested):

mansam (approved):
LGTM!

PR #593, opened 2024-01-16
:ghost: Add questionnare required field to assessment (#588)

ibolton336:
This is needed to aid UI in computing status for an assessment given a questionnaire is marked as not required. If an assessment exists for an archived questionnaire on an app or archetype, the assessment status is no longer valid & the assessment will need to be filtered from status consideration.

https://issues.redhat.com/browse/MTA-1956


mansam:
release-0.3 has been rebased on main so this PR can be closed.

Issue #556, opened 2023-11-14
Upgrade Go 1.20.


jortel:
Upgrade to golang 1.20.
ubi9/go-toolset:latest updated 8 days ago to include 1.20.
This will cause a mass reformat.  In 1.19 gofmt added a number of pedantic rules.
We'll need to update go.mod and re-run `go mod tidy`.

Use _goimports_ for import ordering.

Replace `interface{}` with `any`.

Run gofmt.

Run goimports

PR #595, opened 2024-01-18
:ghost: Go 1.20

jortel:
upgrade Go 1.20.
Replaces _gofmt_ with _goimports_.  Documentation claims it performs the same formatting plus manages/orders imports.
Added missing packages in Makefile.
Updated the go.mod and ran `go mod tidy`.

Simplified controller-gen target.
Updated variable syntax with more traditional.

closes: #556 

NOT FOR 0.3.


mansam (approved):
LGTM

PR #601, opened 2024-02-12
:ghost: Change assessed logic for archetypes to handle zero required 

ibolton336:
- Currently when no required questionnaires exist, archetypes are marked as `assessed:true` which results in additional UI code to rule out this case when calculating status. This PR aims to return `assessed: false` in the case of zero required assessments for an archetype.


mansam (approved):
LGTM

PR #602, opened 2024-02-13
:bug: Return empty list when latest analysis does not exist.

jortel:
Return empty list of issues when no latest analysis. But still 404 when application not found.

https://issues.redhat.com/browse/MTA-2047


mansam (approved):
LGTM

aufi (approved):

PR #603, opened 2024-02-14
:bug: Treat tags that are inherited from archetype assessments as assessment tags rather than archetype tags

mansam:
Prior to this commit, tags that applications inherited due to their archetypes' assessments were categorized as archetype tags rather than assessment tags. This corrects the categorization, and in addition fixes some bugs where tags could be inherited from archived assessments.

Fixes https://issues.redhat.com/browse/MTA-1972

PR #611, opened 2024-02-28
:bug: [Backport][MTA-1972] Treat tags that are inherited from archetype assessments as assessment tags rather than archetype tags (#603)

mansam:
Prior to this commit, tags that applications inherited due to their archetypes' assessments were categorized as archetype tags rather than assessment tags. This corrects the categorization, and in addition fixes some bugs where tags could be inherited from archived assessments.

Fixes https://issues.redhat.com/browse/MTA-1972

### DIFF
--- a/assessment/application.go
+++ b/assessment/application.go
@@ -61,9 +61,7 @@ func (r *ApplicationResolver) Archetypes() (archetypes []Archetype, err error) {
 	return
 }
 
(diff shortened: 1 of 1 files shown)


aufi (approved):
https://github.com/konveyor/tackle2-hub/pull/603

Issue #619, opened 2024-03-28
`builtin.Authenticate` panics if the Authorization header is missing or empty


mansam:
```go
// Authenticate the token
func (r *Builtin) Authenticate(request *Request) (jwToken *jwt.Token, err error) {
	token := strings.Replace(request.Token, "Bearer", "", 1)
	token = strings.Fields(token)[0]
```

If the Authorization header is missing or the string after replacement is nothing but whitespace, then `strings.Fields` will return an empty slice, and the attempt to index into it will panic. This should be fixed to return an auth failure.

jortel:
I do not think changes to the AUTH layer is a  good first issue.

PR #621, opened 2024-03-30
:bug: token authenticate

roceb:
fix for issue / #619 
Missing Bearer token return error instead of panicking. 
Also updated [Client.send() method](https://github.com/konveyor/tackle2-hub/blob/main/binding/client.go#L639) to work with bearer tokens.

### DIFF
--- a/auth/builtin.go
+++ b/auth/builtin.go
@@ -1,6 +1,7 @@
 package auth
 
 import (
+	"errors"
 	"strings"
 
 	"github.com/golang-jwt/jwt/v4"
@@ -21,8 +22,7 @@ type Validator interface {
 }
 
 // NoAuth provider always permits access.
-type NoAuth struct {
-}
+type NoAuth struct{}
 
 // NewToken creates a new signed token.
 func (r NoAuth) NewToken(user string, scopes []string, claims jwt.MapClaims) (signed string, err error) {
@@ -59,13 +59,28 @@ func (r *NoAuth) Refresh(refresh string) (token Token, err error) {
 }
 
 // Builtin auth provider.
-type Builtin struct {
+type Builtin struct{}
+
+// Parse Token out of a string
+func ParseToken(requestToken string) (string, error) {
+	splitToken := strings.Split(requestToken, " ")
+	if splitToken[0] != "Bearer" {
+		return "", errors.New("authentication header not of type bearer")
+	}
+	token := strings.TrimSpace(splitToken[1])
+	if len(token) < 1 {
+		return "", errors.New("no authentication header found")
+	}
+
+	return token, nil
 }
 
 // Authenticate the token
 func (r *Builtin) Authenticate(request *Request) (jwToken *jwt.Token, err error) {
-	token := strings.Replace(request.Token, "Bearer", "", 1)
-	token = strings.Fields(token)[0]
+	token, err := ParseToken(request.Token)
+	if err != nil {
+		return nil, err
+	}
 	defer func() {
 		if err != nil {
 			Log.Info(err.Error())
--- a/auth/builtin.go
+++ b/auth/builtin.go
@@ -1,7 +1,6 @@
 package auth
 
 import (
-	"errors"
 	"strings"
 
 	"github.com/golang-jwt/jwt/v4"
@@ -22,7 +21,8 @@ type Validator interface {
 }
 
 // NoAuth provider always permits access.
-type NoAuth struct{}
+type NoAuth struct {
+}
 
 // NewToken creates a new signed token.
 func (r NoAuth) NewToken(user string, scopes []string, claims jwt.MapClaims) (signed string, err error) {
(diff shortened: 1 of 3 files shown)


auth/builtin.go
```diff
@@ -21,8 +22,7 @@ type Validator interface {
 }
 
 // NoAuth provider always permits access.
-type NoAuth struct {
-}
+type NoAuth struct{}
```
jortel, line 25: Unnecessary reformatting not consistent with rest of the code-base.

auth/builtin.go
```diff
@@ -59,13 +59,28 @@ func (r *NoAuth) Refresh(refresh string) (token Token, err error) {
 }
 
 // Builtin auth provider.
-type Builtin struct {
+type Builtin struct{}
```
jortel, line 62: Unnecessary reformatting not consistent with rest of the code-base.

auth/builtin.go
```diff
@@ -59,13 +59,28 @@ func (r *NoAuth) Refresh(refresh string) (token Token, err error) {
...
+
+// Parse Token out of a string
+func ParseToken(requestToken string) (string, error) {
```
jortel, line 65: Why is this an exported package function instead of a non-exported method on Builtin?

auth/builtin.go
```diff
@@ -59,13 +59,28 @@ func (r *NoAuth) Refresh(refresh string) (token Token, err error) {
...
+	return token, nil
 }
 
 // Authenticate the token
 func (r *Builtin) Authenticate(request *Request) (jwToken *jwt.Token, err error) {
-	token := strings.Replace(request.Token, "Bearer", "", 1)
-	token = strings.Fields(token)[0]
+	token, err := ParseToken(request.Token)
+	if err != nil {
+		return nil, err
```
jortel, line 82: When named return values (pattern is used) the explicit assigned return values here is redundant.  `err` is already assigned and jwtToken is already nil.

jortel: Since parseToken() returns an error, the call should happen after the defer so that it is logged.

auth/builtin.go
```diff
@@ -59,13 +59,28 @@ func (r *NoAuth) Refresh(refresh string) (token Token, err error) {
...
-type Builtin struct {
+type Builtin struct{}
+
+// Parse Token out of a string
+func ParseToken(requestToken string) (string, error) {
+	splitToken := strings.Split(requestToken, " ")
+	if splitToken[0] != "Bearer" {
+		return "", errors.New("authentication header not of type bearer")
+	}
+	token := strings.TrimSpace(splitToken[1])
```
jortel, line 70: This will result in a SEGV when requestToken=`Bearer`. no token.

parseToken() should return `NotValid` which will result in the request http=403 (forbidden) instead of 500.

jortel (changes requested):
I do not think changes to the AUTH layer is a good first issue.

auth/builtin.go
```diff
@@ -59,13 +59,28 @@ func (r *NoAuth) Refresh(refresh string) (token Token, err error) {
 }
 
 // Builtin auth provider.
-type Builtin struct {
+type Builtin struct{}
+
+// Parse Token out of a string
+func ParseToken(requestToken string) (string, error) {
+	splitToken := strings.Split(requestToken, " ")
```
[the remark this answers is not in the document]
roceb: Bearer is for the type of [schema](https://datatracker.ietf.org/doc/html/rfc6750) and JWT is the encoding. JWT encodes to the Bearer schema. So unless there are plans to support different schemas, then Bearer should be included.

As far as:
` requestToken = strings.TrimSpace(requestToken)`
Is there a use case for " Bearer "? This seams like a bad formatted request and shouldn't be allowed.

roceb:
We all have to start somewhere.

auth/builtin.go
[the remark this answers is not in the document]
jortel: > Bearer is for the type of [schema](https://datatracker.ietf.org/doc/html/rfc6750) and JWT is the encoding. JWT encodes to the Bearer schema. So unless there are plans to support different schemas, then Bearer should be included.

Yes, I understand the difference.

> Is there a use case for " Bearer "? This seams like a bad formatted request and shouldn't be allowed.

 Just suggesting that API could be tolerant of curl users trying to do the right thing.
That said, I suppose the standard is not that hard to follow.

jortel:
Thanks for the PR and requested changes @roceb.

We still need to change the [Client.send() method](https://github.com/konveyor/tackle2-hub/blob/main/binding/client.go#L639) like:
```
- request.Header.Set(api.Authorization, r.token.Token)
+ request.Header.Set(api.Authorization, "Bearer "+r.token.Token)
```
Else, the binding won't work with our API.

roceb:
Thank you for your time and patience @jortel . I have updated [Client.send() method](https://github.com/konveyor/tackle2-hub/blob/main/binding/client.go#L639). Let me know if there is anything else?

jortel (approved):
LGTM

PR #623, opened 2024-04-03
:sparkles: Add unassessed risk type and exit early for app risk status calculation

ibolton336:
https://issues.redhat.com/browse/MTA-2308


mansam (changes requested):
Needs the same change for Archetypes but otherwise LGTM

mansam (approved):

PR #635, opened 2024-05-07
:sparkles: Generate Primary Keys.

jortel:
Generate primary keys instead of GORM.
This fixes the issue of GORM reusing the highest key after the model with that ID is deleted.
When the PK is 0, GORM assigns the next (highest) ID.

This approach is to assign the ID ahead of time using a pool managed by tackle.

### DIFF
--- a/database/pk.go
+++ b/database/pk.go
@@ -1,0 +1,160 @@
+package database
+
+import (
+	"errors"
+	"reflect"
+	"strings"
+	"sync"
+
+	"github.com/konveyor/tackle2-hub/model"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+	"gorm.io/gorm/logger"
+)
+
+// PK singleton pk sequence.
+var PK PkSequence
+
+// PkSequence provides a primary key sequence.
+type PkSequence struct {
+	mutex sync.Mutex
+}
+
+// Load highest key for all models.
+func (r *PkSequence) Load(db *gorm.DB, models []any) (err error) {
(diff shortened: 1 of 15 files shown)


migration/v15/model/core.go
```diff
@@ -0,0 +1,314 @@
...
+// Model Base model.
+type Model struct {
+	ID         uint      `gorm:"<-:create;primaryKey"`
+	CreateTime time.Time `gorm:"<-:create;autoCreateTime"`
+	CreateUser string    `gorm:"<-:create"`
+	UpdateUser string
+}
+
+// PK sequence.
+type PK struct {
```
jortel, line 24: Adds.

database/pk.go
```diff
@@ -0,0 +1,160 @@
...
+// Load highest key for all models.
+func (r *PkSequence) Load(db *gorm.DB, models []any) (err error) {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+	for _, m := range models {
+		mt := reflect.TypeOf(m)
+		if mt.Kind() == reflect.Ptr {
+			mt = mt.Elem()
+		}
+		kind := strings.ToUpper(mt.Name())
```
mansam, line 32: This won't work on any tables that have had their name overridden with a TableName method on the model, but the only instance of that we have is on a join table so it should not be relevant. Just pointing it out in the event that changed in the future for some unforeseen reason.

mansam (approved):

mansam (approved):

PR #637, opened 2024-06-04
:bug: Add Token refresh to export/import CLI

aufi:
On slow environments or large data, export/import/clean-all CLI commands might take more than token lifetime is (5 minutes), that failed the operation.

Adding automatic token refresh and doing necessary refactoring to make it possible.

### DIFF
--- a/hack/tool/tackle
+++ b/hack/tool/tackle
@@ -46,6 +46,7 @@ args = parser.parse_args()
 
 EXPORT_MANIFEST_FILENAME = "_manifest"
 KNOWN_CRYPT_STRING = "tackle-cli-known-string-plaintext"
+TOKEN_REFRESH_SECONDS = 240 # 4 minutes to refresh auth token
 
 ###############################################################################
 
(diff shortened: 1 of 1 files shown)


mansam (approved):

hack/tool/tackle
```diff
@@ -299,19 +188,118 @@ class TackleTool:
         print("ERROR: %s record ID %d not found." % (objType, id))
         exit(1)
 
+    def checkTokenLifetime(self):
+        if self.TokenRenewAfter < int(time.time()):
+             self.Token =  getHubToken(self.Url, c['username'], c['password'], False)
+             self.TokenRenewAfter = int(time.time()) + 4*60 # 4 minutes to refresh token
```
shawn-hurley, line 194: Shouldn't the token have this information? If an auth provider has a shorter refresh token, this will fail, I think.

aufi: You're right. The supported go-based implementation also uses Refresh call instead of new Login (https://github.com/konveyor/tackle2-hub/blob/main/binding/client.go#L781).

For this semi/un-supported Python script, that is planned to be replaced with a golang-based tooling, this could be a quickfix IMO.

The time was moved to constant to allow easier adjustment in case of problems.

shawn-hurley: quick fix makes sense, was just a concern, if this is not going to live for much longer I am good :)

PR #645, opened 2024-06-06
:bug: Fix task group update.

jortel:
Fix task group submit.
Update to follow error reporting pattern.

### DIFF
--- a/api/taskgroup.go
+++ b/api/taskgroup.go
@@ -145,17 +145,16 @@ func (h TaskGroupHandler) Create(ctx *gin.Context) {
 		for i := range m.Tasks {
 			task := &tasking.Task{}
 			task.With(&m.Tasks[i])
-			task, err = rtx.TaskManager.Create(h.DB(ctx), task)
+			err = rtx.TaskManager.Create(h.DB(ctx), task)
 			if err != nil {
 				_ = ctx.Error(err)
 				return
 			}
 		}
 	default:
-		h.Respond(ctx,
-			http.StatusBadRequest,
-			gin.H{
-				"error": "state must be ('''|Created|Ready)",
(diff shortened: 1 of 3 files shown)


api/taskgroup.go
```diff
@@ -208,6 +207,16 @@ func (h TaskGroupHandler) Update(ctx *gin.Context) {
...
+		for i := range m.Tasks {
+			task := &m.Tasks[i]
+			if task.ID > 0 {
+				_ = ctx.Error(
+					&BadRequestError{
+						Reason: "already submitted.",
+					})
+				return
+			}
+		}
```
jortel, line 219: hint: Since the update of a _Ready_ task effectively _submits_ it which creates all of the tasks, this check ensure the tasks are not created more than once.

api/taskgroup.go
```diff
@@ -221,17 +230,16 @@ func (h TaskGroupHandler) Update(ctx *gin.Context) {
 		for i := range m.Tasks {
 			task := &tasking.Task{}
 			task.With(&m.Tasks[i])
-			err = rtx.TaskManager.Update(h.DB(ctx), task)
+			err = rtx.TaskManager.Create(h.DB(ctx), task)
```
jortel, line 233: **hint**: main fix.  should be creating the tasks, not updating them.

task/manager.go
```diff
@@ -172,7 +173,7 @@ func (m *Manager) Update(db *gorm.DB, requested *Task) (err error) {
 		if err != nil {
 			return
 		}
-		switch requested.State {
+		switch task.State {
```
jortel, line 176: **hint**: conditional logic should be based on fetched task.

mansam (approved):

PR #647, opened 2024-06-10
:bug: Update export/import CLI no-auth

aufi:
Updating export/import CLI config file parsing to not fail on parsing config file without username or password fields.

Not a blocker, but installations without auth enabled still needed keep username and password fields in config (even with a random content), after this PR they don't have to be there.


aufi:
CI failure is not related

PR #651, opened 2024-06-10
:bug: Pod snapshot best effort when canceled.

jortel:
Pod snapshot can fail when canceling a task with _damaged_ pod.  As a result, the cancel itself should not fail.
Pods usually damaged beyond use when image pull errors.


mansam (approved):
LGTM

PR #655, opened 2024-06-12
:sparkles: Merge task (kind) data.

jortel:
Merge Task (kind) data into Task model data.  This is to support defining data properties in the task kind to be inherrited by models. Properties in Task (model) data have priority.
In the interest of keeping models simple, moves TaskGroup.Propagate() to the api.  No need for it to be in the model. To support this, the Merge() method was also moved to BaseHandler as it is used by multiple handlers.

Example:
```
---
kind: Task
apiVersion: tackle.konveyor.io/v1alpha1
metadata:
  namespace: konveyor-tackle
  name: tech-discovery
spec:
  data:
    mode:
      discovery: true
    rules:
      labels:
        included:
          - techA
          - techB
```

### DIFF
--- a/api/base.go
+++ b/api/base.go
@@ -278,6 +278,38 @@ func (h *BaseHandler) Attachment(ctx *gin.Context, name string) {
 		attachment)
 }
 
+// Merge maps B into A.
+// The B map is the authority.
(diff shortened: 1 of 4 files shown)


mansam (approved):
LGTM

PR #659, opened 2024-06-13
:bug: Fix task filtering and sorting.

jortel:
Fix issues reported [here](https://github.com/konveyor/tackle2-ui/issues/1931#issuecomment-2163782267).
1. sorting seems to work only for `id` field i.e. 
2. filtering not supported for `createUser` and `id` fields( both listed in the enhancement)
3.  application name is missing in the response and in filtering. Filtering by application ID works but this is not so nice.
4. `priority` is missing in the response and in the filtering
5.  for `kind`  the filtering is supported by the server but it seems it's not included in the response

---

Filter and sorting: needed to support GORM join queries which qualifies fields by table.

Example: For a task query joined with application, produces all of the task fields qualified as table.column:
```
SELECT task.id, task.name ...
```
And the (joined) application fields prefixed with `APPLICATION__`
```
SELECT APPLICATION__ID, APPLICATION__NAME ...
```

### DIFF
--- a/api/filter/filter.go
+++ b/api/filter/filter.go
@@ -332,11 +332,16 @@ func (f *Field) Expand() (expanded []Field) {
 // split field name.
 // format: resource.name
 // The resource may be "" (anonymous).
+// The (.) separator is escaped when preceded by (\).
 func (f *Field) split() (relation string, name string) {
 	s := f.Field.Value
 	mark := strings.Index(s, ".")
 	if mark == -1 {
 		name = s
+		return
+	}
+	if mark > 0 && s[mark-1] == '\\' {
+		name = s[:mark-1] + s[mark:]
 		return
 	}
 	relation = s[:mark]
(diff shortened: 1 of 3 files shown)


api/filter/filter.go
```diff
@@ -332,13 +332,18 @@ func (f *Field) Expand() (expanded []Field) {
...
 	s := f.Field.Value
 	mark := strings.Index(s, ".")
 	if mark == -1 {
 		name = s
 		return
 	}
+	if mark > 0 && s[mark-1] == '\\' {
+		name = s[:mark-1] + s[mark:]
+		return
+	}
```
jortel, line 346: Support field names with (.) that are not nested resources (in the filter).
Needed to support GORM join queries which qualifies fields by table.
Example:
For a task query joined with application, all of the task fields are qualified as:
```
SELECT task.id, task.name ...
```
And application fields prefixed with `APPLICATION__`
```
SELECT APPLICATION__ID, APPLICATION__NAME ...
```

mansam (approved):
LGTM

PR #671, opened 2024-06-20
:sparkles: Launch Discovery task when Apps are created/updated

mansam:


api/application.go
```diff
@@ -1333,3 +1346,22 @@ func (r *Stakeholders) contributors() (contributors []model.Stakeholder) {
 	}
 	return
 }
+
+func discover(ctx *gin.Context, application *model.Application) (err error) {
+	rtx := WithContext(ctx)
+	db := rtx.DB.Debug()
```
jortel, line 1352: Isn't the DB in the context already in Debug()?

api/application.go
```diff
@@ -1333,3 +1346,22 @@ func (r *Stakeholders) contributors() (contributors []model.Stakeholder) {
 	}
 	return
 }
+
+func discover(ctx *gin.Context, application *model.Application) (err error) {
```
jortel, line 1350: Let's add:
```
if !Settings.Disconnected {
		return
	}
```
To disable when not connected to a cluster.

importer/manager.go
```diff
@@ -334,11 +341,30 @@ func (m *Manager) createApplication(imp *model.Import) (ok bool) {
...
 		return
 	}
+	// best effort
+	_ = m.discover(app)
 
 	ok = true
 	return
 }
 
+func (m *Manager) discover(application *model.Application) (err error) {
```
jortel, line 351: Let's add:
```
if !Settings.Disconnected {
		return
	}
```
To disable when not connected to a cluster.

jortel (approved):
LGTM!

PR #676, opened 2024-06-26
:sparkles: Reaper optimization.

jortel:
The default reaper frequency is (1) minute.

The reapers are performing 3-4 queries for every _file_ and _bucket_ evaluated for reaping.  With the addition of attaching files to tasks and task-reports, the number of files has increased substantially.  The current algorithm will run 3-4 queries for each resource evaluated.

For example: A deployment with 2000 analyzed applications can have ~25,000 files.  To evaluate 4 resources for references (Task, TaskGroup, Rule, Target) resulted in 100,000 queries each reaper run.

This PR each run (default: 1/ minute) : 
- FileReaper will run exactly **4** queries. 
- BucketReaper will run exactly **3** queries.

### DIFF
--- a/reaper/file.go
+++ b/reaper/file.go
@@ -25,13 +25,25 @@ func (r *FileReaper) Run() {
 		Log.Error(err, "")
 		return
 	}
-	for _, file := range list {
-		busy, err := r.busy(&file)
+	if len(list) == 0 {
+		return
+	}
+	ids := make(map[uint]any)
+	finder := RefFinder{DB: r.DB}
+	for _, m := range []any{
+		&model.Task{},
+		&model.TaskReport{},
+		&model.Rule{},
+		&model.Target{},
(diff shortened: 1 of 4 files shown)


reaper/file.go
```diff
@@ -58,29 +70,6 @@ func (r *FileReaper) Run() {
...
 
-// busy determines if anything references the file.
-func (r *FileReaper) busy(file *model.File) (busy bool, err error) {
-	nRef := int64(0)
-	var n int64
-	ref := RefCounter{DB: r.DB}
-	for _, m := range []any{
-		&model.Task{},
-		&model.TaskReport{},
-		&model.RuleSet{},
```
jortel, line 69: hint: RuleSet removed because it does not have file references.

mansam (approved):
LGTM

Issue #677, opened 2024-06-27
[BUG] Task manager does not detect ImagePullBackoff as an error


dymurray:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

v0.5.0-alpha.1

### Priority

Major

### Current Behavior

Task is in an `ImagePullBackoff` state (due to some sha missing in quay)
```
task-14-smlks                      1/2     ImagePullBackOff   0          6m46s
```

but the task shows as in progress in the UI.

### Expected Behavior

Task should show as errored.

### How Reproducible

Always (Default)

### Steps To Reproduce

Override the image used for analyzer addon and run analysis

### Environment

```markdown
- OS:
```

PR #678, opened 2024-06-27
:bug: Detect image issues.

jortel:
Detect malformed _fqin_ and image pull backoff.  When detected, the task is failed and reported with error and event.

Reported as:
```
events:
    - kind: ImageError
      reason: ImagePullBackOff
      last: 2024-06-27T11:04:14.088026727-07:00
    - kind: PodDeleted
      count: 1
      last: 2024-06-27T11:04:14.126638809-07:00
errors:
    - severity: Error
      description: 'Container (addon) failed: ImagePullBackOff'
```

Detects:
- CrashLoopBackOff,
- ImagePullBackOff
- CreateContainerConfigError
- InvalidImageName
- CreateContainerError

closes: #677

### DIFF
--- a/task/manager.go
+++ b/task/manager.go
@@ -947,10 +947,7 @@ func (m *Manager) podEvent(pod *core.Pod) (events []Event, err error) {
 // podLogs - get and store pod logs as a Files.
 func (m *Manager) podLogs(pod *core.Pod) (files []*model.File, err error) {
 	for _, container := range pod.Status.ContainerStatuses {
-		if container.Started == nil {
-			continue
-		}
-		if !*container.Started {
+		if container.State.Waiting != nil {
 			continue
 		}
 		f, nErr := m.containerLog(pod, container.Name)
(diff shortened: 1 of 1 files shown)


task/manager.go
```diff
@@ -1233,6 +1237,23 @@ func (r *Task) podPending(pod *core.Pod) {
 		status,
 		pod.Status.ContainerStatuses...)
 	for _, status := range status {
+		state := status.State
+		if state.Waiting != nil {
+			waiting := state.Waiting
+			reason := strings.ToLower(waiting.Reason)
+			if strings.Contains(reason, "invalid") || strings.Contains(reason, "backoff") {
```
mansam, line 1244: Would it be better to specifically match `ImagePullBackOff` and `InvalidImageName`, if those are the only cases we intend to catch?

jortel: Perhaps but I cannot find a _const_ in v1 or core packages and concerned about making the match too narrow should the strings change.  Also, a little broader match may catch other similar issues.

mansam (approved):
LGTM

PR #681, opened 2024-07-01
:sparkles: Support task addon succeeded with errors.

jortel:
Support an addon reporting errors independently of reporting state=Failed.


mansam (approved):
LGTM

PR #682, opened 2024-07-02
:bug: cluster refresh atomic.

jortel:
The task Manager.Create() uses the _cluster_ which can have a race condition of being refreshed.
This makes the refresh atomic (by resource kind).


shawn-hurley:
I wonder if using a sync RWMap here might be safer?

shawn-hurley:
I think that once this goes green, this will, to a hopefully large extent help with the KindError inconsistency that is in CI. 

Going to merge once green

jortel:
closed in favor of: #703

PR #699, opened 2024-07-04
:bug: Back-port 0.3: Task.List() no longer include associations; Fix rich context leak.

jortel:
The gin HTTP request handler uses a `gin.Context` pool.  The _RichContext_ contains the response body which is passed to the `Render` middle-ware. 
We need to remove the rich context to prevent holding on to memory while the context is in the pool.

Fixed on main: #710

### DIFF
--- a/api/analysis.go
+++ b/api/analysis.go
@@ -4,6 +4,13 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"io"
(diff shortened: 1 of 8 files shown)


mansam (approved):
LGTM

aufi (approved):

jortel:
0.3 no longer supported.

PR #709, opened 2024-07-05
:seedling: oil the pr ci workflow

djzager:
Needs konveyor/ci#54


jortel (approved):
LGTM but don't know much about actions/CI.

PR #710, opened 2024-07-05
:bug: fix memory leak caused by gin context pooling.

jortel:
The gin HTTP request handler uses a `gin.Context` pool.  The _RichContext_ contains the response body which is passed to the `Render` middle-ware. 
We need to remove the rich context to prevent holding on to memory while the context is in the pool.

_WithContext_ renamed to _RichContext_ for clarity.


mansam (approved):
LGTM

PR #712, opened 2024-07-08
:sparkles: Detect and report pending pod unschedulable.

jortel:
Report event with reason why pending pod cannot be scheduled.
Not sure if this should be merged (or not). Troubleshooting why k8s won't run a pod seems best investigated by snooping around the cluster.  After all ... pods waiting to be scheduled is anticipated (normal) thing on a busy cluster. 
This PR opened only to (potentially) help with troubleshooting broken CI.  It would be better for the CI test to fetch and log task pods stuck at pending.


mansam (approved):
LGTM

PR #715, opened 2024-07-09
:warning: Only trigger discovery tasks if the Application has a repository specified

mansam:


api/application.go
```diff
@@ -388,18 +390,20 @@ func (h ApplicationHandler) Update(ctx *gin.Context) {
...
-			TaskManager: rtx.TaskManager,
-			Client:      rtx.Client,
-			DB:          h.DB(ctx),
-		},
-	}
-	err = tr.Updated(m)
-	if err != nil {
-		_ = ctx.Error(err)
-		return
+	if r.Repository != nil {
```
jortel, line 393: How about moving the check for the repository into the trigger.Application?  When the repository is nil, the trigger could no-op.

trigger/application.go
```diff
@@ -23,14 +25,25 @@ func (r *Application) Updated(m *model.Application) (err error) {
 	if !Settings.Discovery.Enabled {
 		return
 	}
+
+	repo := model.Repository{}
+	err = json.Unmarshal(m.Repository, &repo)
```
mansam, line 30: This will be revisited when the JSON serialization work goes in.

aufi (approved):
LGTM, got this error when tested 0.5.0-beta.1.

PR #724, opened 2024-07-15
:ghost: Update hack add data for identity

sjd78:
Using the hack/add data is convenient way to populate a new dev environment.  The identity script did not create valid identities.  Fields that needed to change to match current expectations have been made.

Now these identities will show up when they're attached to an application.


jortel (approved):
LGTM

Issue #727, opened 2024-07-16
Application create/update don't update associations consistently.


jortel:
Application create/update don't update associations consistently.
They both should omit(clause.Associations)
and replace them.

Likely some other resources need to be _scrubbed_ as well.

PR #733, opened 2024-07-18
:seedling: Omit associations during Create operations

mansam:
When inserting a new record, GORM will also attempt to insert records into tables refered to by many-to-many relationships on the inserted record. This commit attempts to ensure that associations are omitted when inserting records, and then the associations are added to the join tables separately.

Also fixes some spots where the Transaction handler was in the wrong spot in the chain or missing entirely.

Fixes https://github.com/konveyor/tackle2-hub/issues/727

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -1160,6 +1160,7 @@ func (r *Application) With(m *model.Application, tags []model.ApplicationTag) {
 			r.Identities,
 			ref)
 	}
+	r.Tags = []TagRef{}
 	for i := range tags {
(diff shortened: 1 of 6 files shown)


jortel (approved):
LGTM

aufi (approved):
LGTM

Note, we might need test/update export/import CLI (that "nulls" some attributes to avoid circular deps: https://github.com/konveyor/tackle2-hub/blob/main/hack/tool/tackle#L341-L373).

PR #735, opened 2024-07-23
:bug: Revert alpha2 crd

jortel:
Removed v1alpha2.
Adds 0.5 fields to v1alpha1 as optional.
Update the hub to use v1alpah1.
Updates CR status to support conditions.
The addon controller will perform migration to ensure the addon.Container is populated.
The task manager will wait until resources have been reconciled.

---

Moved api/reflect/field.go  Fields() to new _reflect_ package at the root.
Added new functions: 
- HasField()
- Select()
- Omit

My main concern was the the task manager is using Select() to restrict Save() specific fields.  The "Error" fields as incorrect (should be "Errors") and silently didn't update the field.

---

This won't pass CI without operator changes.


k8s/api/tackle/v1alpha1/addon.go
```diff
@@ -43,28 +44,78 @@ type AddonSpec struct {
...
+		if cnd.Type == Ready.Type && cnd.Status == meta.ConditionTrue {
+			ready = true
+			break
+		}
+	}
+	return
+}
+
+// Migrate specification as needed.
+func (r *Addon) Migrate() (updated bool) {
```
jortel, line 87: **hint**: moved this from the controller here.

PR #743, opened 2024-08-09
:sparkles: Upload files instead of multi-part form.

jortel:
Post a _manifest_ file instead of separate multi-part form files.
Much simpler and more easily supports the addon staging the issues and deps files on disk rather than streaming. The more atomic approach will prevent transaction deadlock which can more easily occur when the addon-analyzer builder reported an error (which it should never do).

The uploaded file contains markers used to delimited the documents.
`^]` = `\x1D` = GS (group separator).

```
^]BEGIN-MAIN^]
---
commit: 1234
^]END-MAIN^]
^]BEGIN-ISSUES^]
---
ruleset: ruleset-1
rule: rule-1
incidents:
...
^]END-ISSUES^]
^]BEGIN-DEPS^]
---
name: github.com/jboss
version: 4.0
labels:
- konveyor.io/language=java
- konveyor.io/otherA=dog
^]END-DEPS^]
```

Flow:
1. post (upload) manifest.yaml file.
2. post `ref` to the manifest file.
3. delete manifest file.

Orphaned files will be reaped.

---

The binding client needed to be updated to handle different file encoding (MIME).


mansam (commented):
What changed in the new migration?

mansam (approved):

api/file.go
```diff
@@ -224,27 +184,132 @@ func (h FileHandler) Delete(ctx *gin.Context) {
...
+	}
+	return
+}
 
-	h.Status(ctx, http.StatusNoContent)
+// create a file with request body.
+func (h FileHandler) createBody(ctx *gin.Context, name string) (m *model.File, err error) {
+	m = &model.File{}
+	m.Name = name
+	m.Encoding = binding.MIMEYAML
```
jortel, line 261: **hint**: hard coded to YAML is fine since json is a subset.  Tested and it worked fine.

binding/client.go
```diff
@@ -489,11 +498,19 @@ func (r *Client) FilePost(path, source string, object any) (err error) {
...
+// Returns the created File resource.
+func (r *Client) FilePutEncoded(path, source string, object any, encoding string) (err error) {
 	if source == "" {
 		fields := []Field{
 			{
-				Name:   api.FileField,
-				Reader: bytes.NewReader([]byte{}),
+				Name:     api.FileField,
+				Reader:   bytes.NewReader([]byte{}),
+				Encoding: encoding,
```
mansam, line 513: Does the `Encoding` field need to get set on line 532 as well?

mansam (changes requested):
One question, otherwise looks good.

binding/client.go
jortel: Yes.  Good catch.  Fixed.

mansam (approved):
LGTM

PR #750, opened 2024-09-10
:sparkles: Broker KAI api through hub.

jortel:
Add `/services/` endpoint.
Add `/services/kai/*` endpoint reverse-proxy to route defined in KAI_URL. 
Add auth scopes.

Related: https://github.com/konveyor/operator/pull/376

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v3
         with:
-          go-version: '1.19'
+          go-version: '1.22'
       - run: make fmt
 
   vet:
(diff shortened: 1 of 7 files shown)


aufi (approved):
LGTM

Please correct me if I'm wrong, in "standard" k8s-based Konveyor installation, there are two proxies to access the kai (UI->Hub, Hub->kai) while Hub cares about authorization. The main reason for that is to be able re-use kai service with inside-IDE instalations?

jortel:
> LGTM
> 
> Please correct me if I'm wrong, in "standard" k8s-based Konveyor installation, there are two proxies to access the kai (UI->Hub, Hub->kai) while Hub cares about authorization. The main reason for that is to be able re-use kai service with inside-IDE instalations?

Exactly.  The only purpose of proxying through the hub is to enforce AUTH.

PR #759, opened 2024-10-16
:sparkles: Max DB connections setting.

jortel:
Makes the (max) number of DB connections configurable.  The default=1 which is:
- The safest.
- Most consistent performance (not necessarily the best) on NFS.
- Matches how <=0.5 has always worked.

Increasing max-connections > 1 provides much better throughput on fast filesystems.


mansam (approved):
LGTM

PR #760, opened 2024-10-27
:sparkles: Add GetMultiple function to fetch multiple tasks by IDs

MiriSafra:
Implemented the GetMultiple function in TaskHandler to retrieve multiple tasks based on provided IDs, process the task data, and return a formatted response
related [tackle2-ui PR  #2142 ](https://github.com/konveyor/tackle2-ui/pull/2142)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,6 +141,9 @@ endif
 test:
 	go test -count=1 -v $(shell go list ./... | grep -v "hub/test")
 
+test-db:
+	go test -count=1 -timeout=6h -v ./database...
+
 # Run Hub REST API tests.
 test-api:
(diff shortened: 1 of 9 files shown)


jortel (changes requested):
I don't think this PR is necessary.
The hub /tasks endpoint supports filtering.
Example:  To fetch task 1,2,3
```
GET /tasks?filter=id:(1|2|3)
```
---
The hub API is RESTful and by convention POST is used to create resources. Queries are support using GET with the _filter_ syntax.

MiriSafra:
In light of @jortel feedback, I will proceed to close the PR. Thank you for your input!

PR #761, opened 2024-11-11
:bug: Run task pods with hub UID

jmontleon:


dymurray (approved):

task/manager.go
```diff
@@ -1699,6 +1700,14 @@ func (r *Task) containers(
 			},
 		},
 	}
+	user, err := user.Current()
+	if err != nil {
+		err = liberr.Wrap(err)
```
jortel, line 1705: I appreciate the effort to follow the design pattern of the Task class.

Although the user.Current() and ParseInt() are unlikely to fail, if Current() fails, it will result in a silent SEGV because `user` will be nil.  The error is caught and wrapped but not handled (returned) and `user.Uid` is accessed later.

## A few options

### Option 1:

The intended design pattern here is to build/assemble anything that can fail in Task.Run() and pass those objects into individual (private) methods.  This is for simpler error handling and good functional decomposition.
Task.Run() -> Task.pod() -> Task.specification() -> Task.containers().

Seems best to get/parse the user id in Run() and pass though this call chain instead of changing the return pattern.
Example:
```
func (r *Task) containers(
	addon *crd.Addon,
	extensions []crd.Extension,
	secret *core.Secret,
    uid int64) (init []core.Container, plain []core.Container) {
```
Also seems like better to delegate getting the user to (something like): 
```
Task.getUser() (uid int64, err error){} 
```
because Task.Run() is mainly doing orchestration and even though getting the user only requires 2 steps, it is slightly complex.

### Option 2:

Another strong consideration is to get/parse the UID on hub start up.  The UID is not a per-task specific thing and only needs to be done once.  There are other resources gathered at hub initialization that are deemed essential which the hub deals with by not-starting (panic).  The thinking is that a crippled hub should not run at all and force ENG/cluster-admin to correct.  This could be considered as essential as it would fail to run every task.

To do this, I would recommend adding:

In **settings/hub.go**: 
add `Task.UID int64` and get/parse the user id when loading the settings.  This would make the task UID configurable even though there isn't a clear use case for it but probably not a bad thing. 

For example: This would be use by Task.containers() as:
```
uid = settings.Hub.Task.UID
container.SecurityContext = &core.SecurityContext{
    RunAsUser: &uid,
}
```

settings/hub.go
```diff
@@ -257,6 +260,27 @@ func (r *Hub) Load() (err error) {
...
+		if err != nil {
+			return
+		}
+		r.Task.UID = uid
+	} else {
+		var uid int64
+		var hubUser *user.User
+		hubUser, err = user.Current()
+		if err != nil {
+			return err
```
jortel, line 276: nit: For consistency with _named_ return variables.
```diff
- return err
+ return
```

settings/hub.go
```diff
@@ -257,6 +260,27 @@ func (r *Hub) Load() (err error) {
...
+		}
+		uid, err = strconv.ParseInt(hubUser.Uid, 10, 64)
+		if err != nil {
+			return err
```
jortel, line 280: nit: For consistency with _named_ return variables.
```diff
- return err
+ return
```

jortel (approved):
LGTM
Thanks @jmontleon !

aufi (approved):
LGTM

PR #769, opened 2024-12-08
:sparkles: Add bulk cancel functionality for analyses

MiriSafra:
Added a function for cancelling multiple analyses at once.
Resolves: [tackle2-ui issue #2153]( https://github.com/konveyor/tackle2-ui/issues/2153)

### DIFF
--- a/api/task.go
+++ b/api/task.go
@@ -32,6 +32,7 @@ const (
 	TaskBucketContentRoot    = TaskBucketRoot + "/*" + Wildcard
 	TaskSubmitRoot           = TaskRoot + "/submit"
 	TaskCancelRoot           = TaskRoot + "/cancel"
+	TaskCancelListRoot       = TasksRoot + "/cancel/list"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


jortel (changes requested):
The related issue has not been triaged.
This is a premature optimization and should be closed.

api/task.go
```diff
@@ -487,6 +489,35 @@ func (h TaskHandler) Cancel(ctx *gin.Context) {
 	h.Status(ctx, http.StatusAccepted)
 }
 
+// CancelList godoc
+// @summary Cancel multiple tasks.
+// @description Cancel multiple tasks by IDs.
+// @tags tasks
+// @success 202
+// @router /tasks/cancel/list [put]
```
jortel, line 497: Semantically, I think it would be more correct for the PUT to be against the collection because the action is updating the collection.  Ex: PUT /tasks/cancel (without the list).  The collection implies multiple objects.

jortel (approved):

jortel:
@MiriSafra Thank you for your contribution!

PR #773, opened 2025-01-08
:sparkles: Update Issue Description

aufi:
Issue Description field might be too short to provide enough information about fixing the issue, appending incident Message to it to make it more useful.

This is more a workaround/quickfix, long-term solution is explained in ticket below and at https://github.com/konveyor/analyzer-lsp/discussions/452

Related to: https://issues.redhat.com/browse/MTA-4421 and Hub version of https://github.com/konveyor/kantra/issues/341

### DIFF
--- a/api/analysis.go
+++ b/api/analysis.go
@@ -1109,7 +1109,8 @@ func (h AnalysisHandler) AppIssueReports(ctx *gin.Context) {
 	resources := []*IssueReport{}
 	type M struct {
 		model.Issue
-		Files int
+		Files   int
+		Message string
 	}
 	id := h.pk(ctx)
 	err := h.DB(ctx).First(&model.Application{}, id).Error
@@ -1164,6 +1165,7 @@ func (h AnalysisHandler) AppIssueReports(ctx *gin.Context) {
 		"i.Effort",
 		"i.Labels",
 		"i.Links",
+		"n.Message",
 		"COUNT(distinct n.File) Files")
 	q = q.Table("Issue i,")
 	q = q.Joins("Incident n")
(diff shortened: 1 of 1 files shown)


jortel (approved):
LGTM

api/analysis.go
```diff
@@ -1201,8 +1203,10 @@ func (h AnalysisHandler) AppIssueReports(ctx *gin.Context) {
 	for i := range list {
 		m := list[i]
 		r := &IssueReport{
-			Files:       m.Files,
-			Description: m.Description,
+			Files: m.Files,
+			// Append Incident Message to Description to provide more information on Issue detail
+			// (workaround until Analyzer output Violation struct gets updated to provide better structured data)
+			Description: fmt.Sprintf("%s\n\n%s", m.Description, m.Message),
```
jortel, line 1209: I wonder if this comment is accurate or necessary.  What is the anticipated change in the analyzer data?
This report is a denormalization already.

aufi: I think the main change should be adding Message field to the Violation/Issue, proposal that was discussed mainly with Juanma, Ramon and Shawn is https://github.com/konveyor/analyzer-lsp/discussions/452#discussioncomment-8055352

When implementing the proposal, the Violation struct in analyzer-lsp would be changed (added fields and its population) and components using it as dependency including builder in tackle2-addon-analyzer, likely also Hub and UI will need adapt to this struct change too. That update in Hub should remove change introduced in this PR.

jortel (changes requested):
After further review, including `n.Message` in the result set may greatly change the nature (cardinality) of this report. We also need to consider how pagination will impact this.
 Let's consider this further.

aufi:
(Messages are the same for all Incidents within the Issue)

jortel:
Never mind.  The Group(by) ensures the desired cardinaltiy.

jortel (approved):
LGTM

PR #774, opened 2025-01-17
:sparkles: Support bulk task cancel using filter criteria.

jortel:
Update the bulk (cancel) action route to be an action against the collection.  Also, instead of passing a list of IDs in the body, use a filter instead.  This is more flexible and RESTful.

Example:
/tasks/cancel?filter=id:(1|2|3)
/tasks/cancel?filter=application.id:44
/tasks/cancel?filter=kind:analysis,state=Pending

The only downside is the url length may be an issue for many thousands of cancels (by id).  However, I don't see that as a mainstream use-case.  I would think that use-case would be better supported by filtering by other than id.  To cancel all running tasks, an empty filter may be passed.


dymurray (approved):
I think these changes are a better way to support the feature. LGTM!

PR #776, opened 2025-01-23
:bug: Fix seed dup target labels

jortel:
Fixes an issue with duplicate target labels.
Also, add support for a hub build version needed to detect upgrades. 
The git _tag-hash_ is written to `/etc/hub-build`.  The _abbreviated_ commit hash is appended when HEAD is a commit added to the latest label.
Examples:
```
v0.6.0-beta.1-1-e161635  # tag+hash
v0.6.0-beta.1-1          # tag only
e161635                  # hash only
```

After seeded, the build is stored in settings: `.hub.db.seed.build`.  This is compared with current build to detect when the  build has changed. Seeding is _skipped_ when both the build (version) and the seed data have not changed.

---

The go.mod update was to bump go-utils for better logging of objects.  Included here for better logging of the _Settings_.


seed/target.go
```diff
@@ -87,6 +87,7 @@ func (r *Target) Apply(db *gorm.DB) (err error) {
 		target.Provider = t.Provider
 		target.Choice = t.Choice
 		target.ImageID = f.ID
+		target.Labels = []model.TargetLabel{}
```
jortel, line 90: **hint**: This is the main fix ^.

aufi (approved):
The fix looks good to me, git command in container-build scripts setting hub-build probably needs some update.

```
83.41 git describe --match "v[0-9]*" --tags HEAD > bin/.build
83.41 fatal: No names found, cannot describe anything.
```

jortel:
> The fix looks good to me, git command in container-build scripts setting hub-build probably needs some update.
> 
> ```
> 83.41 git describe --match "v[0-9]*" --tags HEAD > bin/.build
> 83.41 fatal: No names found, cannot describe anything.
> ```

I think it just needs --always flag.

mansam (approved):
LGTM

PR #778, opened 2025-01-28
:sparkles: Use build args for the version string

dymurray:


Dockerfile
```diff
@@ -15,13 +15,15 @@ FROM quay.io/konveyor/static-report as report
 
 FROM registry.access.redhat.com/ubi9/ubi-minimal
 ARG SEED_ROOT
+ARG HUB_VERSION=latest
```
dymurray, line 18: @fbladilo this is where we can just template `CI_VERSION`

jortel: Just a _nit_, but it is the dockerfile for the hub so the HUB_ seems redundant.

Dockerfile
```diff
@@ -15,13 +15,15 @@ FROM quay.io/konveyor/static-report as report
...
 COPY --from=builder /opt/app-root/src/bin/hub /usr/local/bin/tackle-hub
-COPY --from=builder /opt/app-root/src/bin/.build /etc/hub-build
 COPY --from=builder /opt/app-root/src/auth/roles.yaml /tmp/roles.yaml
 COPY --from=builder /opt/app-root/src/auth/users.yaml /tmp/users.yaml
 COPY --from=builder ${SEED_ROOT}/resources/ /tmp/seed
 COPY --from=report /usr/local/static-report /tmp/analysis/report
 
+RUN echo "${HUB_VERSION}" >> /etc/hub-build
```
rayfordj, line 25: Is this append (`>>`) desired instead of the overwrite (`>`) that was [previously used](https://github.com/konveyor/tackle2-hub/pull/778/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L63) to generate this file?  Is there any concern where appending could introduce a problem that wouldn't be observed by overwriting?

rayfordj: It seems safe given it should be a new file creation regardless, but wanted to toss it out there... :wink:

rayfordj (approved):
/lgtm

Dockerfile
jortel: Agreed `>` would be more correct.

dymurray: Fair point; considering the base is always reset I don't think this would be an issue but no reason to append!

PR #782, opened 2025-02-11
:sparkles: Support extenion compat with addon as regex.

jortel:
Both the analyzer and tech-discovery both use the same providers. An extension having compatibility with multiple addons is not a common/mainstream use case. To get around this, we have defined a duplicate set of extensions for each provider. One with compatibility with the `analyzer` addon and the other for the `tech-discovery` addon.

After a recent refactor of the generic extension into multiple extensions (go, python, nodejs), it became obvious that maintain the duplicate set of extensions is painful and likely error prone.

It seems a better approach is to change `Extention.Spec.Addon` to (optionally) define a regex. This approach caters to the mainstream use case that extensions have compatibility with one addon.  It also avoids making the field an _array_ requiring a CRD change.  This is which is MUCH simpler.

The existing extensions will need to be updated to be compatible with both addons.
Example (java extension):
```
addon: ^(analyzer|tech-discovery)$
```

The `discovery-` extensions would be deleted.

### DIFF
--- a/settings/hub.go
+++ b/settings/hub.go
@@ -368,6 +368,7 @@ func (r *Hub) namespace() (ns string, err error) {
 // build returns the hub build version.
 // This is expected to be the output of `git describe`.
 // Examples:
+//
 //	v0.6.0-ea89gcd
 //	v0.6.0
 func (r *Hub) build() (version string, err error) {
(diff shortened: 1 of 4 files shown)


aufi (approved):
Looks great :+1:

PR #789, opened 2025-03-17
:bug: Improved application LIST performance.

jortel:
Currently, the application List endpoint performs a query for each application to fetch the tags.  This does not scale well.  This PR fetches all of the tags into a map keyed by application ID.  Pretty sure this must be similar behavior to the un-correlated subquery done by GORM when pre-loading.

The ApplicationTag (join) models are small so not worried about memory footprint.

This reduced the performance (on my minikube with SSD) from 1.8s => ~700ms with 5k applications which is a 60% performance increase. I would expect a more significant impact on cluster hdd backed file systems.

Roughly related to: https://issues.redhat.com/browse/MTA-4165

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -1107,8 +1107,8 @@ func (h *ApplicationHandler) tagMap(
 	applications []model.Application) (mp map[uint][]model.ApplicationTag, err error) {
 	ids := []uint{}
 	for i := range applications {
-		app := &applications[i]
-		ids = append(ids, app.ID)
+		m := &applications[i]
+		ids = append(ids, m.ID)
 	}
 	mp = make(map[uint][]model.ApplicationTag)
 	list := []model.ApplicationTag{}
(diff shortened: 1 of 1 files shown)


api/application.go
```diff
@@ -1108,6 +1101,28 @@ func (h ApplicationHandler) AssessmentCreate(ctx *gin.Context) {
...
+	err = db.Find(&list, "ApplicationID", ids).Error
+	if err != nil {
+		return
+	}
+	for _, tag := range list {
+		mp[tag.ApplicationID] = append(mp[tag.ApplicationID], tag)
+	}
+	return
+}
+
```
jortel, line 1125: Re: https://github.com/konveyor/tackle2-hub/blob/main/api/application.go#L175  This omits the:
```
h.preLoad(h.DB(ctx), clause.Associations)
```
I don't see why we'd need the associations.  In fact, it seem like it would load a ton of unwanted data.

jortel: The tag needs to be fetched for the ref.name.
This can be optimized though to prevent fetching the application.

mansam (approved):
LGTM

PR #794, opened 2025-04-15
:bug: Fix error in addon: unknown user: 1001.

jortel:
The binding should not be loading the settings.

The addon adapter should only be loading the addon settings.


aufi (approved):
Thanks for removing dependency on `settings`, LGTM.

`PrepareRichclient()` method is typically used outside of this repo (e.g. in API or analysis tests), so no other repos should be affected with this change :+1:

PR #795, opened 2025-04-22
:ghost: Update golang min version to 1.21

fbladilo:
- The hub project is a go 1.20 but the slices library was only included in the standard library starting with go 1.21. 
This discrepancy causes Hermeto (Cachi2) to fail fetching libraries. The minimum go version should be 1.21.

Refs:

https://pkg.go.dev/slices?tab=versions

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/konveyor/tackle2-hub
 
-go 1.20
+go 1.21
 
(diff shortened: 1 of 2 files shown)


aufi (approved):
LGTM

jortel (approved):
lgtm

PR #796, opened 2025-04-23
:seedling: MTA-5006 Update auth setup after Keycloak update

aufi:
There few more mandatory User fields to be setup when auth is enabled, adding those. Or we might update the realm setup to not require it by default.

Fixes: https://issues.redhat.com/browse/MTA-5006


aufi:
Successfuly tested locally, to have working API right after the installation, ensure `Settings.Keycloak.RequirePasswordUpdate` is false.

jortel (approved):
lgtm

PR #798, opened 2025-04-28
:sparkles: Encrypted data enhancements.

jortel:
### The Goal 

The goal is to better support the concept of any model containing secret (encrypted) fields.

### Details
Upgrade encryption to AES+GCM.  Stronger and errors better when incorrect key is used to decrypt.

Add `Secret` object used to encrypt/decrypt a string or fields in a struct or a map[string]any. This centralizes encryption functionality and supports upcoming needs for other models to have encrypted fields.  Prevents double encryption naively. 

_encryption_ package renamed: _secret_.
with API for convenience:

- secret.Use(passphrase string) 
- secret.Encrypt(object any)
- secret.Decrypt(object any)

Migration v17 to re-encrypt identities with stronger cipher.

Introduces the `secret:""` tag.  Fields to be encrypted/decrypted using secret.Encrypt(m) and secret.Decrypt(m).

Example(model):
```
// Identity represents and identity with a set of credentials.
type Identity struct {
	Model
	Kind         string `gorm:"not null"`
	Name         string `gorm:"index;unique;not null"`
	Description  string
	User         string
	Password     string        `secret:""`
	Key          string        `secret:""`
	Settings     string        `secret:""`
	Proxies      []Proxy       `gorm:"constraint:OnDelete:SET NULL"`
	Applications []Application `gorm:"many2many:ApplicationIdentity;constraint:OnDelete:CASCADE"`
}
```


mansam (approved):
LGTM

PR #802, opened 2025-05-02
:bug: Updates PK catalog when resources has explicitly assigned ID.

jortel:
Updates PK catalog when resources has explicitly assigned ID.
For example when populated for demos using scripts.
Example:

TagCategory:
```
id: 14
name: "Test
```
Tag:
```
id:65
categoryID: 14
```

Unit test updated.


aufi (approved):

mansam (approved):
LGTM

PR #818, opened 2025-05-13
:sparkles: Add asset generator.

jortel:


aufi (approved):
From a quick look, looks good.

PR #819, opened 2025-05-17
:sparkles: TaskGroup modes.

jortel:
Update the TaskGroup to support both batch and piipeline modes.

Mode=**Batch** (current behavior) provides a convenient method for creating homogeneous (kind, addon, extensions) tasks. This is used by the UI to do bulk analysis.
Mode=**Pipeline** provides a method for creating a group of non-homogeneous tasks that are executed in order. **The driving use case** for this is: the discovery tasks.  They need to run in-order.  Further, the selection of addon and extensions for tech-discovery needs to happen after lang-discovery so that the appropriate extension is selected. This approach is much cleaner than the dependency rules that causes the tech-discovery to be Postponed until lang-discovery has completed. It solves the discovery problem because the trigger can create and submit the tasks a a (pipeline) group.  When submitted, the manager will set only the first task state=Ready.  After it completes, the next task will be set state=Ready (and so on).  Since addon and extension selection happens when the task becomes Ready, the Language= tag will have been set by the lang-discovery.


reaper/task.go
```diff
@@ -225,25 +261,36 @@ func (r *GroupReaper) Run() {
 				r.delete(m)
 			}
 		case task.Ready:
-			if len(m.Tasks) == 0 {
-				r.delete(m)
-				continue
-			}
-			if m.HasBucket() {
+			mark := m.CreateTime
+			if time.Since(mark) > time.Hour {
```
jortel, line 265: hint: Let's leave this around for an hour to support troubleshooting.

reaper/task.go
```diff
@@ -225,25 +261,36 @@ func (r *GroupReaper) Run() {
...
+	nChanged := 0
+	if m.HasBucket() {
 		Log.Info("Group bucket released.", "id", m.ID)
-	} else {
-		Log.Error(err, "")
+		m.SetBucket(nil)
+		nChanged++
+	}
+	if len(m.List) > 0 {
+		m.List = nil
```
jortel, line 286: hint: Clear the list to reduce storage. After the group has been submitted, the tasks in the List have been created as _real_ tasks.

aufi (approved):
From quick look, looks good.

Thanks for updating Task README with Modes!

PR #822, opened 2025-05-26
:sparkles: Support API rendering of large collections with MUCH less memory.

jortel:
Moves SAX-like encoders to from analysis.go to context.go for more general usage.  
Using these encoders instead of standard json|yaml encoders for large arrays GREATLY reduce memory footprint.
Add (cursor) _Iterator_ to better support queries using joins instead of pre-loading.  The iterator groups rows (by ID) into batches. Joins return a flattened object graph. Each batch is passed to the supplied _Builder_ function to construct the resource.

Note:
_In Java, SAX stands for Simple API for XML.
It is a stream-based, event-driven API used for parsing XML documents. Unlike DOM parsers, which load the entire XML document into memory as a tree, SAX parsers read the XML document sequentially and trigger events as elements are encountered, making them faster and more memory-efficient, especially for large XML files._

### DIFF
--- a/api/context.go
+++ b/api/context.go
@@ -113,94 +113,108 @@ func Transaction(ctx *gin.Context) {
 // Render renders the response based on the Accept: header.
 // Opinionated towards json.
 func Render() gin.HandlerFunc {
-	return func(ctx *gin.Context) {
-		ctx.Next()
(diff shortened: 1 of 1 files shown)


aufi (approved):
LGTM

PR #892, opened 2025-08-26
:bug: fix assessment risk and confidence when listing applications.

### LINKED ISSUE #891: [BUG] Assessed application do not show the correct assessment risk
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Konveyor version

main

### Priority

Critical

### Current Behavior

When assessing an application the risk type is set as `Unknown` in the application drawer as well as in the reports tab

### Expected Behavior

_No response_

### How Reproducible

Always (Default)

### Steps To Reproduce

1. Create an application
2. perform an assessment with risk of type high for example

### Environment

```markdown
- OS:
```

### Anything else?

In the reports tab under the `Identified risks` section, it is possible to filter the questions by the risk and it will show the relvant applications that should have that risk status.

jortel:
The _resolvers_ expect the `Application.Assessments[].Sections` to be populated.
This works for Get() because it uses db.Fetch() with clause._Associations_ which fetches the entire nested Assessment object.  List() uses a a custom join for performance and memory footprint. 

This is a regression introduced by https://github.com/konveyor/tackle2-hub/pull/789 RE: application list performance and memory footprint.  OPTIMIZATION.

closes #891

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -188,24 +188,25 @@ func (h ApplicationHandler) List(ctx *gin.Context) {
 
 	type M struct {
 		*model.Application
-		IdentityId      uint
-		IdentityName    string
-		ServiceId       uint
-		ServiceName     string
-		OwnerId         uint
-		OwnerName       string
-		ContributorId   uint
-		ContributorName string
-		WaveId          uint
-		WaveName        string
-		PlatformId      uint
-		PlatformName    string
-		ReviewId        uint
-		AssessmentId    uint
-		ManifestId      uint
-		QuestionnaireId uint
-		AnalysisId      uint
-		Effort          int
+		IdentityId         uint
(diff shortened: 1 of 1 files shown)


api/application.go
```diff
@@ -188,24 +188,25 @@ func (h ApplicationHandler) List(ctx *gin.Context) {
...
+		OwnerName          string
+		ContributorId      uint
+		ContributorName    string
+		WaveId             uint
+		WaveName           string
+		PlatformId         uint
+		PlatformName       string
+		ReviewId           uint
+		AssessmentId       uint
+		AssessmentSections []byte
```
jortel, line 205: **hint**: just `AssessmentSections` added.

dymurray (approved):

PR #896, opened 2025-08-28
:bug: Fix ambiguous field filter/sort problem on task list

sjd78:
Resolves: https://issues.redhat.com/browse/MTA-5739

Hub needed an extra field rename on the Task list function to disambiguate the "kind" field to come from the Task table.


jortel (approved):
lgtm

PR #922, opened 2025-09-29
:sparkles: Postgres PoC.

jortel:
## Goal
Postgres proof-of-concept.
Part of a performance and scale effort. My theory is that the EOF experienced by the addons (binding client) is caused by contention for the single DB connection. PG will support concurrent reader/writer.  The goal is to port the code to work with PG.  No other way to determine the LOE and challenges.  Then, measure performance with 5k applications.

### Main porting changes/notes:
- PG is case insensitive and all identifiers (table, column names etc).  The GORM `NamingStrategy` is not ubiquitous enough to deal with this which resulted in issues with the Migrator.  As a result, **all of the identifiers changed to lower snake case and plural table names**.  This is the recommended path of least resistance.
- PG is stricter SQL.
  - subqueries must be aliased: SELECT * from (?) _alias_.
  - All selected columns not being aggregated, must be in the group by.
- PG `user` and `binary` are keywords.  Identity.User and Application.Binary renamed.
- PG locking with regard to mixing of ALTER TABLE and SELECTs done by the GORM Migrator means we cannot run AutoMigrate() in a transaction without deadlock.
- Sorting:
  - ERROR: collations are not supported by type timestamp with time zone (SQLSTATE 42804)
  - ERROR: collation "nocase" for encoding "UTF8" does not exist (SQLSTATE 42704)'
  - Updated the api/sort to no longer use the sqlite NOCASE for now.  Need a solution for this.

### Good side effects:
- PG auto-increment based on sequences and does not re-use IDs.
- RH engineering can (potentially) connect to the DB externally for troubleshooting.
- PG has better error reporting.  Especially for constraint violations.
- Better support for backup/restore.

TODO:
- [ ] Migrate all of the data sqlite => PG.  [pgloader](https://pgloader.io) is the popular choice but would need to evaluate performance over just copying the models internally.  Mainly due to the identifier mapping.
- [ ] Identify problems of running AutoMigrate() outside of a transaction.  Data migration could still be in a Tx.
- [ ] Find a solution for NOCASE (case insensitive sorting).
- [ ] Evaluate impact of multiple writers.  For example: primary key assignment is likely broken and need to switch to auto-assignment using sequences.


api/application.go
```diff
@@ -194,7 +194,7 @@ func (h ApplicationHandler) List(ctx *gin.Context) {
 		_ = ctx.Error(err)
 		return
 	}
-	filter = filter.Renamed("platform.id", "PlatformId")
+	filter = filter.Renamed("platform.id", "platform_id")
```
aufi: Curious about the field names change to snake-case, is this something want to do to have better readable/focusing on some conventions, or support 2 databases are forcing us to do that?

PR #973, opened 2025-12-09
:ghost: Refactor hub into internal/ and shared/ packages.

### LINKED ISSUE #971: Decouple hub.binding.
Currently the addons and now Kantra are using the binding.  However, because the binding references api.* which references model.* and other packages.  Also, the hub is 1 module and as a result, to use the binding, go-mod drags in the entire hub and all of it's dependencies.

The plan is to use sub-modules:
Add a new _shared_ module with the bare-bones REST api structs.
This will require the api package to import from resources and refactor the concrete resources to be based on these structs then add the With() and Model() methods.

The binding package would then import from resources instead of api.

This should be considered for the addon package as well.

jortel:
Moved hub (internal) functionality into the new `internal/` package.  As of Go 1.4 the special _internal_ package is not exported.  As a bonus, this eliminates the need for the tackle2-addon repository all together.

The goal is to decouple the binding and addon (adapter) from the main hub module to reduce go-mod dependencies bloat when imported.  Packages refactored to separate exported types and consts from hub internals having dependencies:

The `scm` package has a counterpart in both `internal` and `shared/addon` each to facilitate usages in the respective environments.  The internal/scm also adds local mirroring.

The `command` package has as counterpart in the  `shared/addon` each to facilitate usages in the addon environment.  Mainly for reporting/logging.

The `api` package refactored:
- The `internal/api/resource` contains definitions based on `api` that defines With() and Model() methods.
- The handlers updated to alias resource types defined in `api/resource`.

The _shared_ (exported) packages have been moved to a `shared/` submodule.  This includes:
- settings
- **api**
- **binding**
- **addon** (adapter)
- scm (source code management)
- ssh
- tar
- env
- task

This requires wide-spread import changes.
 
closes #971 

Ref: https://go.dev/doc/modules/layout

### DIFF
--- a/shared/addon/adapter/adapter.go
+++ b/shared/addon/adapter/adapter.go
@@ -33,6 +33,7 @@ func init() {
 
 // Client
 type Client = binding.Client
+type RichClient = binding.RichClient
 type Params = binding.Params
 type Param = binding.Param
 type Path = binding.Path
@@ -70,8 +71,6 @@ type Adapter struct {
 	Task
 	// Log API.
 	Log logapi.Logger
-	// Client binding client.
-	Client *binding.RichClient
 	// Wrap error API.
 	Wrap func(error, ...any) error
 	//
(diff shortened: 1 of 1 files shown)


jortel:
@coderabbitai This is a large refactor where the majority of the code in [list key new paths or packages] has been moved from [old paths/packages] without changes. Please treat those sections as relocated existing code and prioritize reviewing only the actual modifications, new files, or dependency updates.

env/pkg.go
```diff
@@ -0,0 +1,45 @@
+package env
```
aufi: This package is in top-level directory since it is used by multiple packages or there is other reason it is important to be placed right there?

aufi (approved):
Gave it a quick look (compared to number of changed lines), overall looks well and make a good sense. I'm okay with merging, just leaving few clarification questions.

Tiny, non-blocking, tiny question is on naming imports like `json2` (for internal/migration/json) or `model2`, just thinking if there is some more descriptive name/alias (looking that it is related to Konveyor, not a golang stdlib new version of e.g. json).

PR #1042, opened 2026-04-29
:sparkles: builtin OIDC provider.

jortel:
Removes all dependencies on Keycloak.
The `builtin` provider rewritten to BE An OIDC provider based on Zitadel lib.
Adds:
- RBAC inventory (user, role, permission)
- Basic-Auth mostly to support tools and testing (only valid for hub user - not federated).
- Support for PAT (personal access tokens).  AKA: api-key.
- Support for device access grants (DAG) for CLI/tools.
- Support client registration and client-credentials grants.
- Includes basic login web pages to be replaced with proper javascript pages designed by UXD.
- Support optional federation to external IdP.  (Think Keycloak).
- Support optional LDAP (backend) integration with group/role mapping.

Packages updated:
- internal/auth - Rewritten auth layer.
- internal/api - AuthHandler provides endpoints for oidc/ and RBAC and Token inventory.
- internal/database - just added OpenTest() to better support unit tests.
- internal/migration - add v22 with model changes.
- internal/model - links v22.
- internal/task - updated to use api-keys.
- internal/k8s/api - add (2) new CRD for: _OpenidProvider_ and _LdapProvider_.
- shared/api - added route (const) and REST resources for model.
- shared/addon - adapter changes related to auth and binding.
- shared/binding - add support for new auth flows and REST resources.
- test - add/updated integration tests.

MANY other packages files updated due to needed upgrade to Go 1.25.

Re: [Enhancment](https://github.com/konveyor/enhancements/blob/master/enhancements/hub-oidc-provider/README.md).


aufi (approved):
This PR is really an OIDC re-implementation inside the Hub codebase, it feels to me to be little too much integrated as an integral part of the Hub. I'm not sure if this was required, I'd expected having it more separated for better maintenance, etc. But just my $0.02, not a blocker.

## Other my comments

Looking on HTML fragment template in internal/auth/storage.go, this might be adjustable (e.g. for d/s), but could be updated later.

Thanks for tests/README.md, maybe there could be something in https://github.com/konveyor/tackle2-hub/tree/main/docs too.

## Other comments (from AI) to consider (up to your decision)

🚨 GORM queries using camelCase instead of DB column names
In internal/auth/storage.go, there are several queries formatted like this:

First(m, "ClientId", clientId) on line 62

First(grant, "refreshToken", digest) on line 401

First(m, "authId", id) on line 1055

Where("refreshToken", ...) on line 1091

First(m, "authId", authId) on line 1199

Issue: GORM expects actual database column names, which are typically snake_case (e.g., client_id, refresh_token, auth_id). On Postgres, querying ClientId will fail because it looks for a non-existent clientid column.
Recommendation: Explicitly use the correct DB column names (e.g., client_id = ?, refresh_token = ?, auth_id = ?) or use map/struct conditions.

🔒 Security Blocker: User API leaks password hashes
In internal/api/resource/user.go:17, the response object maps r.Password = m.Password, which is then exposed to the outside in UserGet/UserList (internal/api/auth.go:440-477). Even though it is a bcrypt hash, it should never be returned in a REST response.
Recommendation: The password field should strictly be input-only and omitted from the With() serialization.

🐛 Bug: RoleUpdate breaks the permissions cache
After calling Association("Permissions").Replace(m.Permissions), RoleUpdate saves m directly into the cache (internal/api/auth.go:740-746). However, resource.Role.Model() populates permissions using only their IDs (internal/api/resource/role.go:27-32), while Role.GetScopes() relies on p.Scope (internal/auth/cache/pkg.go:51-54).
Result: After a role update, the cache will contain permissions missing their Scope data. This means users with that role might lose their actual permissions until a cache refresh or restart occurs.
Recommendation: Reload the role with its associations after the Replace call, similar to how it is handled in RoleCreate.

⚠️ Security/Design: Token validation lacks issuer/audience checks
The RSA/HMAC token parsing uses jwt.WithoutClaimsValidation(), and the custom validToken() only checks sub, scope, and exp (internal/auth/builtin.go). I don't see any validation for iss (issuer) or aud (audience).
Recommendation: If the hub is acting as both the OIDC provider and the resource server, we should at least have a conscious decision (or tests) ensuring that an access token minted for a different client/audience cannot be used to call the Hub API.

⚠️ Security: op.WithAllowInsecure() is hardcoded
In internal/auth/builtin.go, the OIDC provider is created with op.WithAllowInsecure(). I understand this might be necessary for local/dev HTTP testing, but for production, this needs to be feature-flagged or restricted by configuration. Otherwise, it's too easy to accidentally run the OIDC provider in an insecure mode.

jortel:
RE: GORM and column case. The comment is based on the assumption that we're using Postgres (which is case sensitive and snake case is best).  The hub uses sqlite3 which is not case sensitive and configures GROM for camel case.

jortel:
Added (`iss`) issuer verification.

jortel:
Fixed permission issue in the cache when roles updated.

jortel:
Made TLS skip verify configurable with optional CA support.

jmontleon:
Right now on the test environment if you log in with keycloak, then log out of MTA, you are returned to the login prompt. But your IdP session is still valid and so you (or anyone who wants to come by and impersonate you) can click Sign in with keycloak and get back into MTA unprompted.

It is possible to modify the behavior, at the very least with keycloak, to end the IdP session as well so that the behavior resembles what a user probably expects.

PR #1051, opened 2026-06-05
:bug: End upstream Keycloak session on logout.

### LINKED ISSUE #1050: [BUG] Logging out of MTA when using Keycloak to login to MTA after OIDC update doesn't end the keycloak session
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Konveyor version

main

### Priority

Major

### Current Behavior

Right now on the test environment if you log in with keycloak, then log out of MTA, you are returned to the login prompt. But your IdP session is still valid and so you (or anyone who wants to come by and impersonate you) can click Sign in with keycloak and get back into MTA unprompted.

It is possible to modify the behavior, at the very least with keycloak, to end the IdP session as well so that the behavior resembles what a user probably expects.

### Expected Behavior

_No response_

### How Reproducible

Always (Default)

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
```

### Anything else?

_No response_

jmontleon:
Implement CanTerminateSessionFromRequest on Storage so that logging out of the hub also terminates the upstream Keycloak session for federated OIDC users. The redirect chains through Keycloak's end_session_endpoint before returning to the client's post_logout_redirect_uri.

Non-federated (local/LDAP) users are unaffected; their logout redirects directly as before.

Keycloak configuration: in the Keycloak admin console, the hub's client must have a "Valid post logout redirect URIs" entry that matches the URI the client sends in post_logout_redirect_uri (e.g. the UI's origin). Without this, Keycloak will ignore the redirect and show its own logged-out page.

closes #1050

## Summary by CodeRabbit

* **New Features**
  * Added federated identity provider logout support for session termination
  * Session end-session requests now properly redirect to upstream identity providers when configured

* **Tests**
  * Added comprehensive test coverage for federated logout URL handling and session termination flows


jmontleon:
@coderabbitai full review

internal/auth/idp.go
```diff
@@ -126,6 +127,38 @@ func (h *FedIdpHandler) identitySaved(m *Identity) {
 	h.cache.IdentitySaved(m)
 }
 
+// EndSessionURL builds the upstream IdP's end_session URL.
+// The postLogoutRedirectURI is passed through to the IdP so the user
+// is redirected back to the original destination after the IdP session ends.
+func (h *FedIdpHandler) EndSessionURL(postLogoutRedirectURI string) (logoutURL string, ok bool, err error) {
```
jortel: Returning the `ok` along with the url and error could be simpler.  As is it represents 3 states which is a little confusing:
1. failed: returned an error
2.  no error but ok=false (no idea what 'ok' means).
3. no error and ok=true (again, no idea what 'ok' means).

Recommend eliminating the returned 'ok' value.  When err (nil), simpler for the caller to just determine what to do based on the returned URL != "".

Note: Although, `ok` is common in Go it's avoided the hub as it has no clear semantic meaning.  A value with semantic meaning is strongly preferred.

internal/auth/idp.go
```diff
@@ -126,6 +127,38 @@ func (h *FedIdpHandler) identitySaved(m *Identity) {
...
+	if !federated.Idp.Enabled {
+		return
+	}
+	var rpClient rp.RelyingParty
+	rpClient, err = h.RpClient()
+	if err != nil || rpClient == nil {
```
jortel: Following Go convention, h.RpClient() returns a valid pointer else an error.  While the defensive checking for nil isn't _bad_, but it suggests a non-standard convention.

internal/auth/idp.go
```diff
@@ -126,6 +127,38 @@ func (h *FedIdpHandler) identitySaved(m *Identity) {
...
+		return
+	}
+	endpoint := rpClient.GetEndSessionEndpoint()
+	if endpoint == "" {
+		return
+	}
+	parsed, pErr := url.Parse(endpoint)
```
jortel: Recommend:
```
parsed, err := url.Parse(endpoint)
```
Direct assignment to `err` simpler without shadowing issues.

Good job 👍  wrapping the error.

internal/auth/idp_test.go
```diff
@@ -0,0 +1,189 @@
+package auth
```
jortel: The convention in hub packages is to put all of the unit tests in `pkg_test.go`.  There isn't much advantage to putting these tests in their own file.  The goal is to reduce the number of files in the package for readability. 
Please move these tests to pkg_test.go.

internal/auth/storage.go
```diff
@@ -407,6 +407,45 @@ func (r *Storage) TerminateSession(ctx context.Context, userID, _ string) (err e
...
+func (r *Storage) TerminateSessionFromRequest(
+	ctx context.Context,
+	req *op.EndSessionRequest) (redirect string, err error) {
+	//
+	err = r.TerminateSession(ctx, req.UserID, req.ClientID)
+	if err != nil {
+		return
+	}
+	redirect = req.RedirectURI
+	if r.idpHandler == nil || req.UserID == "" {
```
jortel: Not _bad_ to test for a nil handler but Storage guaranteed to be constructed with one.  This would be the only place in the code where it's checked.  Again, not _bad_ just not consistent.

internal/auth/storage.go
```diff
@@ -407,6 +407,45 @@ func (r *Storage) TerminateSession(ctx context.Context, userID, _ string) (err e
...
+		return
+	}
+	s, sErr := r.findSubject(req.UserID)
```
jortel: Recommend:
```
s, err = r.findSubject(req.UserID)
```
Direct assignment to `err` simpler.

internal/auth/storage.go
```diff
@@ -407,6 +407,45 @@ func (r *Storage) TerminateSession(ctx context.Context, userID, _ string) (err e
...
+		}
+		return
+	}
+	if s == nil {
+		return
+	}
+	if !s.IsIdentity() || s.Identity.Kind != IdentityKindOpenid {
+		return
+	}
+	logoutURL, ok, eErr := r.idpHandler.EndSessionURL(req.RedirectURI)
```
jortel: See other comments re: `ok` and direct assignment to `err`.

jortel (changes requested):
@jmontleon Thanks for the PR!
I tested in my AWS cluster and worked well.  I agree this is better behavior.
Just a few nits.

internal/auth/storage.go
```diff
@@ -407,6 +407,41 @@ func (r *Storage) TerminateSession(ctx context.Context, userID, _ string) (err e
...
+	redirect = req.RedirectURI
+	if req.UserID == "" {
+		return
+	}
+	s, err := r.findSubject(req.UserID)
+	if err != nil {
+		if errors.Is(err, &NotFound{}) {
+			err = nil
+		}
+		return
```
jmontleon, line 430: This looks funny to me, but based upon the explanation it seems reasonable, "If we return with a NotFound error still set, the Zitadel OP treats it as a session termination failure and returns an error response to the client. But "subject not found" during logout isn't a
  failure — it just means we can't determine whether the user is federated, so we should return successfully with the default redirect.
  
  Without err = nil, a user who was already cleaned up from the cache/DB would get a logout error instead of a clean redirect."

jortel (approved):
LGTM

internal/auth/storage.go
jortel: Implied in the direct assignment suggestion was to err = nil when NotFound.

PR #1089, opened 2026-07-01
:sparkles: Add login page support

### LINKED ISSUE #3397: Add login pages.
Currently the hub has home-grown login pages:
- Login
- Device access grant.

The Login page needs to be parameterized (template) so the hub can inject. 

Current embedded pages:
- [login](https://github.com/konveyor/tackle2-hub/blob/main/internal/auth/storage.go#L893)
- [Device Auth](https://github.com/konveyor/tackle2-hub/blob/main/internal/auth/dag.go#L53)

sjd78:
Resolves: https://github.com/konveyor/tackle2-ui/issues/3397

Create a `login-page` project to handle the various required OIDC pages.  It provides:
  - User login page
  - Device verification page
  - Device success page
  - Login session expiration page

The project is nodejs based, uses rspack to build the static assets, and is based on the PatternFly 6 login page component.

Changes to the hub code to support the login page are minimal. The hub code will now stream the login page assets to the client as built by the `login-page` project. When serving the primary `index.html` page, a configuration JSON object is constructed and embedded in the page. This configuration object is used to determine which page to render.

Branding is supported via a `branding` directory with a `strings.json` file that contains the branding strings. The `login-page` project reads this file at build time and makes it available to the components as a global constant. Changing the branding requires rebuilding the `login-page` project with a `BRANDING` environment variable set to the desired branding asset directory.

Assisted-by: Cursor:claude-opus-4.6-high

## Summary by CodeRabbit

* **New Features**
  * Added a modern branded login experience with support for user login, federated identity providers, device authorization, and session-expired pages.
  * Added configurable branding, including titles, imagery, backgrounds, favicons, and theme styles.
  * Login-page assets are now built and included with the hub application.
  * Added local development scripts for running the hub and UI together.

* **Documentation**
  * Added setup, development, frontend architecture, and branding guidance.

* **Bug Fixes**
  * Improved authentication error and expired-session handling.
  * Expanded supported OAuth/OIDC grants, redirect URLs, and scopes.


login-page/src/index.html
```diff
@@ -0,0 +1,23 @@
...
+    The hub replaces the contents of this script tag before serving the page.
+    It injects a JSON object matching the LoginConfig interface so the React
+    app knows which page to render and what dynamic values to use (form
+    actions, error messages, federated IdP config).
+
+    The placeholder value is intentionally incomplete JSON so that a failure
+    to inject is immediately visible rather than silently falling back.
+  -->
+  <script id="login-config">
+    window.__LOGIN_CONFIG__ = null;
```
rszwajko: 1. the statement is subject to code formatting and does not match the pattern in `page.go`
2. consider adding the entire tag with JSON inside i.e. `<script id="login-config" type="application/json">...`  or replacing a simple placeholder string i.e. `window.__LOGIN_CONFIG__ =  SOME_PLACEHOLDER`;
3. do we need to `JSON.parse()` the injected data?

internal/frontend/auth/content/branding/logo.svg
rszwajko, line 1: Should we use the Konveyor or Tackle here? If Konveyor then we can re-use  https://github.com/konveyor/tackle2-ui/blob/main/branding/images/masthead-logo.svg

internal/loginpage/page.go
```diff
@@ -0,0 +1,76 @@
...
+		return
+	}
+
+	configJSON, err := json.Marshal(cfg)
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+
+	injected := "window.__LOGIN_CONFIG__=" + string(configJSON)
```
rszwajko: do we need some HTML escaping/sanitization here?

rszwajko (commented):
Looks good! check my comments below

internal/frontend/auth/content/branding/logo.svg
dymurray: Konveyor IMO

internal/loginpage/page.go
sjd78: No ... this should be using the standard go type to JSON transform used everywhere else.

login-page/src/index.html
sjd78: Yeah the template matching setup should be better.  Ideally the index.html can just have a template string embedded.  I plan on iterating on this approach a bit.

sjd78: Switched template matching to go templates.

sjd78: This has been changed to use a go template and no longer minifiy the index page, much safer.

internal/frontend/auth/content/branding/logo.svg
sjd78: Using the konveyor logo and text now.

dymurray:
I'll leave Jeff to review the code; but tested the entire login flow on latest commit with zero issues :+1:

internal/auth/storage.go
```diff
@@ -813,8 +816,8 @@ func (r *Login) parseCredentials() (err error) {
 		err = liberr.Wrap(err)
 		return
 	}
-	r.login = r.request.PostFormValue("login")
-	r.password = r.request.PostFormValue("password")
+	r.login = r.request.PostFormValue("pf-login-username-id")
+	r.password = r.request.PostFormValue("pf-login-password-id")
 	return
 }
```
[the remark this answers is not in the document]
sjd78: Already fixed.

internal/frontend/auth/content/src/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <title>Tackle Hub</title>
```
rszwajko, line 6: can we use branding here?

internal/frontend/auth/content/src/SessionExpiredPage.tsx
```diff
@@ -0,0 +1,50 @@
...
+  return (
+    <PFLoginPage
+      loginTitle="Session Expired"
+      backgroundImgSrc={styles?.backgroundImage ?? ""}
+      brandImgProps={{
+        src: styles?.brandImage ?? "",
+        alt: application?.name ?? application?.title ?? "",
+        widths: { default: "260px" },
+      }}
+      textContent=""
```
rszwajko, line 36: we should add this prop to branding and provide some default text about Konveyor for upstream. Without that text the right side of the login page looks a bit empty (only Konveyor logo is there).

internal/frontend/auth/content/src/types.ts
```diff
@@ -0,0 +1,30 @@
+// LoginConfig is injected by the hub as window.__LOGIN_CONFIG__ before serving the
+// page. It carries only per-request dynamic values; branding (titles, logos,
+// background images) is baked into the bundle at build time.
+export interface LoginConfig {
+  // Selects which page component to render.
+  page: "login" | "device-verify" | "device-success" | "session-expired";
+
+  // Login page: POST target URL including authRequestId query param.
+  formAction?: string;
```
rszwajko, line 9: @jortel 
do you handle gracefully a case where old `authRequestId` is sent?
User can leave the the login page open for a long time.

internal/frontend/auth/content/src/UserLoginPage.tsx
```diff
@@ -0,0 +1,77 @@
...
+        widths: { default: "260px" }
+      }}
+      textContent={loginPage?.textContent ?? ""}
+      socialMediaLoginContent={socialLoginContent}
+      socialMediaLoginAriaLabel={
+        federatedIdp ? "Alternative login options" : undefined
+      }
+      footerListVariants={ListVariant.inline}
+    >
+      <LoginForm
```
rszwajko, line 56: Consider (visually) highlighting the main login area. The proposal from https://www.patternfly.org/components/login-page/design-guidelines  looks good to me:
1. main area is wrapped in some kind of card
2. the background (including the right side panel with extra content) has different color
3. background image is not necessary but if we have some art work that can be re-used then great

rszwajko (approved):
Works for me! See my comments for possible follow-ups (not blockers).

internal/frontend/auth/content/src/types.ts
dymurray: It would be nice to have it autoredirect back to login or something

sjd78: - Commit 185d2b2 will renew an old authRequestId
- Commit 731b72b adds a session-expired screen for an unknown authRequestId

There is no way to recover the incoming params that allow pushing a user back to the application if the authRequestId is unknown.  Best we can do is say "login session is expired".  I've had this exact thing happen with Red Hat SSO with a page sitting on login for too long.

internal/frontend/auth/content/src/index.html
sjd78: Yeah there are a few places to add branding support.  The page title and the favicon are both part of that.

internal/frontend/auth/content/src/types.ts
sjd78: FWIW - restarting the hub process / pod will drop all existing authRequestIds and cause every open login screen to hit the session-expired screen

jortel (changes requested):
https://github.com/sjd78/tackle2-hub/pull/1

internal/frontend/auth/content/src/SessionExpiredPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+        icon={ExclamationCircleIcon}
+        status="danger"
+        titleText="Session Expired"
+        headingLevel="h2"
+      >
+        <EmptyStateBody>
+          Login session has expired.
+        </EmptyStateBody>
+        <EmptyStateActions>
+          <Button component="a" href="/" variant="primary">
```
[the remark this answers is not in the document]
sjd78: In this case it is not an issue.

internal/frontend/auth/content/src/index.html
```diff
@@ -0,0 +1,14 @@
...
+  <script id="login-config">
+    window.__LOGIN_CONFIG__ = {{ . }};
```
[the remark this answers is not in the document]
sjd78: The value coming in is coming from a go type struct marshalled to json, so this should not be a problem

jortel (approved):
LGMT

PR #160, opened 2022-10-10
[backport-2.1.2] Avoid cyclic dependencies

aufi:
v2.1.2 backport of https://github.com/konveyor/tackle2-hub/pull/152

Adding GORM BeforeCreate Hook for Dependencies to avoid creating cyclic Application Dependencies.

Example: having dependencies apps 1->2, 2->3, then 3->1 should not be allowed, but 1->3 is still OK.

Related to https://issues.redhat.com/browse/TACKLE-814

Signed-off-by: Marek Aufart [maufart@redhat.com](mailto:maufart@redhat.com)

PR #1119, opened 2026-08-10
:sparkles: Add Agent endpoints.

### LINKED ISSUE #1112: :sparkles: Add agent endpoints.
Add agentic-controller endpoints. 

 ## Routes

  - **`agentic/agents`** — Agent
    - Actions:
      - list
      - get
      - create
      - update
      - delete
  - **`agentic/skills`** — SkillCard
    - Actions:
      - list
      - get
      - create
      - update
      - delete
  - **`agentic/skillcollections`** — SkillCollection
    - Actions:
      - list
      - get
      - create
      - update
      - delete
  - **`agentic/gateways`** — Gateway
    - Actions:
      - list
      - get
      - create
      - update
      - delete
  - **`agentic/agentruns`** — AgentRun
    - Actions:
      - list
      - get
      - create
  - **`agentic/runs/:name/cancel`** — AgentRun
    - Actions:
      - cancel
  - **`agentic/agentruns/:name/acp/nonce`** — AgentRun nonce
    - Actions:
      - post
  - **`agentic/agentruns/:name/acp?nonce=123`** — AgentRun ACP connection
    - Actions:
      - websocket proxy to agent pod ACP endpoint
  - **`agentic/workflows`** — AgentWorkflow
    - Actions:
      - list
      - get
      - create
      - update
      - delete
  - **`agentic/workflowruns`** — AgentWorkflowRun
    - Actions:
      - list
      - get
      - create
  - **`agentic/workflowruns/:name/cancel`** — AgentWorkflowRun
    - Actions:
      - cancel

## Roles/Scopes

| Resource | admin | architect | migrator | project-manager |
  | --- | --- | --- | --- | --- |
  | agentic.agentruns | `*` | `*` | `*` | `get` |
  | agentic.agentruns.acp | `*` | `*` | `*` | `get` |
  | agentic.agents | `*` | `get` | `get` | `get` |
  | agentic.gateways | `*` | `get` | `get` | `get` |
  | agentic.skillcollections | `*` | `*` | `get` | `get` |
  | agentic.skills | `*` | `*` | `get` | `get` |
  | agentic.workflowruns | `*` | `*` | `*` | `get` |
  | agentic.workflows | `*` | `*` | `get` | `get` |

  ## Key behaviors

  - **All reads are request-driven** — `client.List()`/`client.Get()` with no informer cache, since agent resource counts are small (tens, not thousands).
  - **Agent/AgentWorkflow listing** filters by `konveyor.io/managed=true` label for the UI. Other resource types are listed unfiltered.
  - **AgentRun/AgentWorkflowRun creation** mints a scoped API token (with `AddonScopes`), stores it in a Secret, and injects `HUB_BASE_URL`, `HUB_APP_ID`, and the
  token Secret into `spec.env`/`spec.envFrom` on the CR before creating it.
  - **Cancel** revokes the Hub API token and sets `spec.cancel: true` on the CR.
  - **ACP WebSocket proxy** at `agent/runs/:name/acp` — Hub resolves the Sandbox Service DNS, reads the ACP secret key from the Secret referenced by
  `status.secretKeyRef`, injects the `X-Secret-Key` header, and proxies WebSocket frames bidirectionally between the UI and the agent pod.

jortel:
closes #1112 

## Summary by CodeRabbit

* **New Features**
  * Added an Agentic API for managing agents, skills, skill collections, gateways, workflows, and runs.
  * Added real-time agent connections through authenticated WebSocket sessions.
  * Added Kubernetes resource integration and automatic run credential handling.
  * Added role-based access for agentic resources and an agent service account.
* **Bug Fixes**
  * Improved API error responses for unavailable and missing resources.
* **Security**
  * Added single-use, expiring authentication nonces for agent connections.

### DIFF
--- a/hack/manifest.yaml
+++ b/hack/manifest.yaml
@@ -73,7 +73,7 @@ metadata:
   namespace: tackle-hub
   name:  tackle-hub
 rules:
-- apiGroups: ["", "batch", "tackle.konveyor.io"]
+- apiGroups: ["", "batch", "tackle.konveyor.io", "konveyor.io"]
   resources: ["*"]
   verbs: ["*"]
 
--- a/internal/api/agentic.go
+++ b/internal/api/agentic.go
@@ -73,14 +73,14 @@ func (h AgenticHandler) AddRoutes(e *gin.Engine) {
 	// AgentRun
 	routeGroup = e.Group("/")
 	routeGroup.Use(Required("agentic.runs"))
-	routeGroup.GET(api.AgenticAgentRunsRoute, h.RunList)
-	routeGroup.GET(api.AgenticAgentRunsRoute+"/", h.RunList)
-	routeGroup.POST(api.AgenticAgentRunsRoute, h.RunCreate)
-	routeGroup.GET(api.AgenticAgentRunRoute, h.RunGet)
+	routeGroup.GET(api.AgenticAgentRunsRoute, h.AgentRunList)
+	routeGroup.GET(api.AgenticAgentRunsRoute+"/", h.AgentRunList)
+	routeGroup.POST(api.AgenticAgentRunsRoute, h.AgentRunCreate)
+	routeGroup.GET(api.AgenticAgentRunRoute, h.AgentRunGet)
 	// AgentRun ACP
 	routeGroup = e.Group("/")
 	routeGroup.Use(Required("agentic.runs.acp"))
-	routeGroup.GET(api.AgenticAgentRunACPRoute, h.RunACP)
+	routeGroup.GET(api.AgenticAgentRunACPRoute, h.AgentRunACP)
 	// AgentWorkflow
 	routeGroup = e.Group("/")
 	routeGroup.Use(Required("agentic.workflows"))
@@ -663,7 +663,7 @@ func (h AgenticHandler) GatewayDelete(ctx *gin.Context) {
 // AgentRun
 //
 
-// RunGet godoc
+// AgentRunGet godoc
 // @summary Get an agent run by name.
 // @description Get an agent run by name.
 // @tags runs
@@ -671,7 +671,7 @@ func (h AgenticHandler) GatewayDelete(ctx *gin.Context) {
 // @success 200 {object} AgentRun
 // @router /agentic/runs/{name} [get]
 // @param name path string true "AgentRun name"
-func (h AgenticHandler) RunGet(ctx *gin.Context) {
+func (h AgenticHandler) AgentRunGet(ctx *gin.Context) {
 	r := &AgentRun{}
 	err := h.Client(ctx).Get(
 		context.TODO(),
@@ -687,14 +687,14 @@ func (h AgenticHandler) RunGet(ctx *gin.Context) {
 	h.Respond(ctx, http.StatusOK, r)
 }
 
-// RunList godoc
+// AgentRunList godoc
 // @summary List all agent runs.
 // @description List all agent runs.
 // @tags runs
 // @produce json
 // @success 200 {object} []AgentRun
 // @router /agentic/runs [get]
-func (h AgenticHandler) RunList(ctx *gin.Context) {
+func (h AgenticHandler) AgentRunList(ctx *gin.Context) {
 	list := &AgentRunList{}
 	err := h.Client(ctx).List(
 		context.TODO(),
@@ -710,7 +710,7 @@ func (h AgenticHandler) RunList(ctx *gin.Context) {
 	h.Respond(ctx, http.StatusOK, list.Items)
 }
 
-// RunCreate godoc
+// AgentRunCreate godoc
 // @summary Create an agent run.
 // @description Create an agent run.
 // @tags runs
@@ -719,7 +719,7 @@ func (h AgenticHandler) RunList(ctx *gin.Context) {
 // @success 201 {object} AgentRun
 // @router /agentic/runs [post]
 // @param run body AgentRun true "AgentRun data"
-func (h AgenticHandler) RunCreate(ctx *gin.Context) {
+func (h AgenticHandler) AgentRunCreate(ctx *gin.Context) {
 	r := &AgentRun{}
 	err := h.Bind(ctx, r)
 	if err != nil {
@@ -783,14 +783,14 @@ func (h AgenticHandler) RunCreate(ctx *gin.Context) {
 	h.Respond(ctx, http.StatusCreated, r)
 }
 
-// RunACP godoc
+// AgentRunACP godoc
 // @summary WebSocket proxy to agent pod ACP endpoint.
 // @description Upgrades the connection to a WebSocket and proxies frames
 // @description bidirectionally to the agent pod's ACP endpoint.
 // @tags runs
 // @router /agentic/runs/{name}/acp [get]
 // @param name path string true "AgentRun name"
-func (h AgenticHandler) RunACP(ctx *gin.Context) {
+func (h AgenticHandler) AgentRunACP(ctx *gin.Context) {
 	r := &AgentRun{}
 	err := h.Client(ctx).Get(
 		context.TODO(),
--- a/internal/api/agentic.go
+++ b/internal/api/agentic.go
@@ -74,15 +74,15 @@ func (h *AgenticHandler) AddRoutes(e *gin.Engine) {
 	routeGroup.DELETE(api.AgenticGatewayRoute, h.GatewayDelete)
 	// AgentRun
 	routeGroup = e.Group("/")
-	routeGroup.Use(Required("agentic.runs"))
+	routeGroup.Use(Required("agentic.agentruns"))
 	routeGroup.GET(api.AgenticAgentRunsRoute, h.AgentRunList)
 	routeGroup.GET(api.AgenticAgentRunsRoute+"/", h.AgentRunList)
 	routeGroup.POST(api.AgenticAgentRunsRoute, h.AgentRunCreate)
 	routeGroup.GET(api.AgenticAgentRunRoute, h.AgentRunGet)
 	// AgentRun ACP
 	routeGroup = e.Group("/")
-	routeGroup.Use(Required("agentic.runs.acp"))
-	routeGroup.POST(api.AgenticAgentRunConnRoute, h.AgentRunConnectNonce)
+	routeGroup.Use(Required("agentic.agentruns.acp"))
+	routeGroup.POST(api.AgenticAgentRunConnNonceRoute, h.AgentRunConnectNonce)
 	e.GET(api.AgenticAgentRunConnRoute, h.AgentRunConnect)
 	// AgentWorkflow
 	routeGroup = e.Group("/")
--- a/internal/api/agentic.go
+++ b/internal/api/agentic.go
@@ -722,7 +722,11 @@ func (h AgenticHandler) AgentRunCreate(ctx *gin.Context) {
 		_ = ctx.Error(err)
 		return
 	}
-	kind := h.Kind(ctx, r)
+	kind, err := h.Kind(ctx, r)
+	if err != nil {
+		_ = ctx.Error(err)
+		return
+	}
 	r.SetGroupVersionKind(agent.GroupVersion.WithKind(kind))
 	client := h.Client(ctx)
 	secret, tokenId, err := h.tokenSecret(r)
@@ -1061,7 +1065,11 @@ func (h *AgenticHandler) WorkflowRunCreate(ctx *gin.Context) {
 		_ = ctx.Error(err)
 		return
 	}
-	kind := h.Kind(ctx, r)
+	kind, err := h.Kind(ctx, r)
+	if err != nil {
+		_ = ctx.Error(err)
+		return
+	}
 	r.SetGroupVersionKind(agent.GroupVersion.WithKind(kind))
 	client := h.Client(ctx)
 	secret, tokenId, err := h.tokenSecret(r)
--- a/internal/api/agentic.go
+++ b/internal/api/agentic.go
@@ -668,7 +668,7 @@ func (h AgenticHandler) GatewayDelete(ctx *gin.Context) {
 // @tags runs
 // @produce json
 // @success 200 {object} AgentRun
-// @router /agentic/runs/{name} [get]
+// @router /agentic/agentruns/{name} [get]
 // @param name path string true "AgentRun name"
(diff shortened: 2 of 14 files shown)


internal/api/agentic.go
```diff
@@ -0,0 +1,1329 @@
...
+		_ = ctx.Error(err)
+		return
+	}
+	agentCon := AgentConn{
+		ctx:     ctx,
+		sandbox: r.Status.SandboxName,
+		key:     key,
+	}
+	agentCon.Relay()
+}
```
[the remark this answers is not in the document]
jortel: Associating the nonce to a specific run does not add value because the scope (the nonce represents) is for run resource and not a specific instance.

internal/api/error.go
```diff
@@ -103,6 +104,22 @@ func (r *TrackerError) Is(err error) (matched bool) {
...
+}
+
+func (r *NotAvailableError) Error() string {
+	return r.Reason
+}
+
+func (r *NotAvailableError) Is(err error) (matched bool) {
+	var target *NotAvailableError
+	matched = errors.As(err, &target)
+	return
```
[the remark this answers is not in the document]
jortel: Looks like the contract changed in Go 1.18.  I'll do a follow up in the entire codebase.

jortel:
@coderabbitai review

hack/manifest.yaml
```diff
@@ -73,7 +73,7 @@ metadata:
   namespace: tackle-hub
   name:  tackle-hub
 rules:
-- apiGroups: ["", "batch", "tackle.konveyor.io"]
+- apiGroups: ["", "batch", "tackle.konveyor.io", "konveyor.io"]
   resources: ["*"]
   verbs: ["*"]
```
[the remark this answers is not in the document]
jortel: This is just "hack" file. hack is not released code.

internal/auth/seed/serviceaccounts.yaml
```diff
@@ -3,3 +3,8 @@
   name: task.addon
   roles:
     - addon
+- id: 10
+  description: Agent harness.
+  name: agent.harness
```
ibolton336: This seeds `agent.harness`, but `tokenSecret` looks up `agentic.harness` (`internal/api/agentic.go:1165`). With auth on, every run create dies before it starts:

```
POST /agentic/agentruns -> 404 {"error":"SA (agentic.harness) not-found"}
```

Renaming the seed entry to `agentic.harness`, nothing else changed, flips that same request to **201** — AgentRun CR and token Secret both land in the namespace.

Worth knowing there's no API workaround to fall back on: creating the SA by hand with `POST /serviceaccounts` panics in `auth/cache` `addSaScopes` when the role is passed by name — the failed `roleById` lookup shadows `r`, so `r.ID` on the not-found path is a nil deref. That's pre-existing on `main`, not this PR; I can file it separately if useful.

I have this rename plus the role grants sitting in jortel/tackle2-hub#2. The roles half is now redundant with 8d62107, so it's down to the one-line rename if you want to take it that way.

internal/auth/seed/roles.yaml
```diff
@@ -404,6 +452,30 @@
     - name: adoptionplans
       verbs:
         - post
+    - name: agentic.agentruns
+      verbs:
+        - get
+    - name: agentic.agentruns.acp
```
ibolton336: `agentic.agentruns.acp` is a POST-only route (the nonce mint), so `get` grants project-manager nothing:

```
POST /agentic/agentruns/{name}/acp/nonce
-> 403 Required scope not granted: agentic.agentruns.acp:POST
```

If the intent is that a project-manager can watch a run but not drive it, the scope model can't express that today — `acp` is a single bit over a bidirectional channel, and the grant that lets you watch is the same grant that lets you steer once steering comes off the feature flag. Probably worth settling before TP which way this goes: drop the entry as dead, or split watch from steer.

internal/api/agentic.go
```diff
@@ -0,0 +1,1333 @@
...
+// @description Upgrades the connection to a WebSocket and proxies frames
+// @description bidirectionally to the agent pod's ACP endpoint. Authenticated
+// @description by a nonce query parameter obtained from POST /agentic/runs/{name}/acp.
+// @tags runs
+// @router /agentic/agentruns/{name}/acp [get]
+// @param name path string true "AgentRun name"
+// @param nonce query string true "Single-use nonce"
+func (h AgenticHandler) AgentRunConnect(ctx *gin.Context) {
+	nonce := ctx.Query("nonce")
+	err := h.nonce.Redeem(nonce)
```
ibolton336, line 828: `Redeem` consumes the nonce before the readiness checks below, so a pre-flight failure burns it:

```
mint nonce -> connect (status not yet populated) -> 503 "Status not populated."
replay the same nonce                            -> 401
```

That's the common path rather than an edge case — sandbox pods have no readiness gate, so `SandboxName`/`SecretKeyRef` are routinely empty for the first seconds of a run and a UI that opens the console on run start polls straight into it. Each retry then costs a fresh nonce round-trip.

Redeeming first is the right instinct — it keeps an unauthenticated caller from probing run existence and status through this route — so the cheap fix is to validate here and not delete until the upstream dial succeeds.

internal/api/agentic.go
```diff
@@ -0,0 +1,1333 @@
...
+	defer func() {
+		_ = client.Close()
+	}()
+	header := http.Header{
+		"X-Secret-Key": {r.key},
+	}
+	remote, _, err := websocket.DefaultDialer.Dial(
+		u,
+		header)
+	if err != nil {
```
ibolton336, line 1241: By this point `Upgrade` has already hijacked the client connection, so returning here closes it with no close frame: the browser sees a successful handshake followed by a bare disconnect, and can't distinguish "sandbox isn't listening yet" from "run is gone". A `WriteControl(websocket.CloseMessage, websocket.FormatCloseMessage(...))` before returning would give the UI something to branch on, and pairs with the nonce point above — those two together are what a client needs to retry sanely.

ibolton336 (commented):
Ran this branch on an auth-on rig (`AUTH_REQUIRED=true`, image built from 9e0fbc4, fresh DB) and probed the agentic surface as admin / architect / migrator / project-manager / addon.

Verified working:

- the role grants from 8d62107 land — architect and migrator get 200/201 across list and nonce mint, migrator correctly 403s on `POST /agentic/agents`, project-manager sees run history
- unauthenticated 401 everywhere; replayed and garbage nonces 401
- the GVK fix in e8b8d1e works: the ownerRef is now `konveyor.io/v1alpha1` (was a bare `v1alpha1`), and deleting an AgentRun garbage-collects its token Secret — confirmed in-cluster

Four notes inline. The first one blocks every run create under auth.

internal/auth/seed/roles.yaml
```diff
@@ -19,6 +19,30 @@
...
         - post
+    - name: agentic.agentruns
+      verbs:
+        - '*'
+    - name: agentic.agentruns.acp
+      verbs:
+        - '*'
+    - name: agentic.agents
+      verbs:
+        - get
```
ibolton336, line 30: One more from the rig, on the same commit: architect is `get`-only on `agentic.agents` and `agentic.gateways`, so defining an agent or a gateway is admin-only.

```
architect POST /agentic/agents   -> 403 Required scope not granted: agentic.agents:POST
architect POST /agentic/gateways -> 403 Required scope not granted: agentic.gateways:POST
architect POST /agentic/workflows -> passes the scope check
```

That's a coherent position — Agent and Gateway are closer to platform config than to migration work, and an architect composing workflows out of admin-provided agents is a reasonable split. Flagging it because it's load-bearing elsewhere: the management screens on the tackle2-ui side (konveyor/tackle2-ui#3504) put agent and gateway authoring in the architect persona, so if this is the intended model I need to gate those screens on admin rather than architect. Which way do you want it?

jortel:
@coderabbitai What is my current review limit. explain.

jortel:
@coderabbitai What is my plan?

internal/auth/seed/serviceaccounts.yaml
jortel: > Renaming the seed entry to agentic.harness, nothing else changed, flips that same request to 201 — AgentRun CR and token Secret both land in the namespace.

Fixed.

> Worth knowing there's no API workaround to fall back on: creating the SA by hand with POST /serviceaccounts panics in auth/cache addSaScopes when the role is passed by name — the failed roleById lookup shadows r, so r.ID on the not-found path is a nil deref. That's pre-existing on main, not this PR; I can file it separately if useful.

The SA role refs must populate the ID.  The name in ALL refs is read-only.
However, the 500 is an issued missed by tests and I will fix in separate PR.

https://github.com/konveyor/tackle2-hub/issues/1124

internal/auth/seed/roles.yaml
jortel: Fixed.

internal/api/agentic.go
jortel: A nonce by definition can be used once.  Also, the nonce is only good for 30 seconds so the UI will need to renew it as some point depending on how long it takes for the controller to reconcile the AgentRun and the the pod scheduler to run the pod.

Since there is a try/retry flow here, I wonder if the endpoint should support an (optional) parameter that specifies the number of retries and/or the retry interval?

jortel:
@coderabbitai review

jortel:
@coderabbitai review

internal/api/agentic.go
jortel: Fixing.

jortel:
@coderabbitai review

internal/api/agentic.go
```diff
@@ -0,0 +1,1432 @@
...
+	}
+}
+
+// sendClose writes a WebSocket close frame with the given code and reason.
+func (r *AgentConn) sendClose(conn *websocket.Conn, code int, reason string) {
+	_ = conn.WriteControl(
+		websocket.CloseMessage,
+		websocket.FormatCloseMessage(code, reason),
+		time.Now().Add(time.Second))
+}
```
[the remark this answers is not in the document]
jortel: Fixed.

aufi (approved):
Overall LGTM, pasting two comments from claude review to be considered.

  ### 1. WebSocket Origin Validation (Security)

  **File:** `internal/api/agentic.go:1273`

  The WebSocket upgrader accepts connections from any origin:

  ```go
  upgrader := websocket.Upgrader{
      CheckOrigin: func(_ *http.Request) bool {
          return true  // Accepts all origins - CSRF vulnerability
      },
  }

  Recommendation: Implement origin whitelist or at minimum validate against Settings.Hub.URL to prevent CSRF attacks.

  2. Missing Error Wrapping

  File: internal/api/agentic.go (multiple locations)

  Errors from external k8s client are not wrapped with liberr.Wrap(). According to project standards (CLAUDE.md), errors should be wrapped "at the edge" when returned by external packages.

  Example in acpKey() (~line 1198):
  err = h.Client(ctx).Get(context.TODO(), k8s.ObjectKey{...}, secret)
  if err != nil {
      return  // Missing: err = liberr.Wrap(err)
  }

  This pattern appears in multiple methods: AgentRunCreate, WorkflowRunCreate, tokenSecret, etc.

  Recommendation: Add err = liberr.Wrap(err) for all k8s client operations before returning.
  ```

ibolton336:
@jortel Re-tested your newest push (quay `:agent` @ `bb01527d`, built from head `cd72328`) live on our ROKS cluster — four console-driven runs with the browser WebSocket instrumented: clean baseline, graceful pod kill mid-turn, tab-close abandon, and force-kill.

**The close-frame relay work holds up in all four.** Every socket teardown ended in a real close frame with a usable code — `1001 "harness shutting down"` at natural end, `1000` on the pod kills, `4404 "run not available"` / `1013 "sandbox not ready"` on post-exit redials (closed in 5–111 ms). Nothing the UI has to special-case as a bare disconnect. The 45 s post-exit hold I hit on the previous image (`d849875b` — a redial catching the endpoint-less Service and hanging until i/o timeout) did not reproduce, so the sandbox dial-timeout idea demotes to nice-to-have hardening. And with agentic-controller #160's readiness gate live, the UI dials once and opens in 116 ms — the multi-attempt dial dance is gone, so no retry parameter is needed on the hub side.

Three things worth a look before this merges:

1. **`ErrCloseSent` log noise** — every relay teardown logs `websocket: close sent` with a stack trace from `sendClose` (the second goroutine's duplicate close attempt). It fires on every normal teardown, so worth ignoring `ErrCloseSent` (or demoting it below error) to keep real relay errors visible.
2. **Harness token self-revocation 403s on every run** — teardown logs `POST /auth/tokens/{id}/revoke` → `403 "Required scope not granted: tokens:POST"`. The harness role still lacks the `tokens` scope (the open question from the seed round), so minted run tokens are never revoked and just age out on TTL. Fine as a follow-up if you'd rather not touch seeds again in this PR — but it fires on literally every run.
3. **The one review question still open** ([thread](https://github.com/konveyor/tackle2-hub/pull/1119#discussion_r3777828975)): architect remains `get`-only on `agentic.agents` and `agentic.gateways`, so agent/gateway authoring is admin-only. Intended? The answer decides which personas get the management screens in the console.

Separately — you mentioned you had no way to test agent runs locally. The rig all of these findings came out of is now committed in my client repo: [docs/hub-auth-rig.md](https://github.com/ibolton336/agentcontroller-client/blob/main/docs/hub-auth-rig.md). `hack/hub-auth-up.sh` stands up an auth-on hub (no operator, no keycloak, disposable minikube profile) in ~40 s, and the "testing a hub branch" loop builds your working tree straight into the profile's docker daemon — `HUB_LOCAL_BUILD=1`, no registry round-trip, ~6 min per iteration. `hack/hub-auth-probe.sh` then runs the 21-leg role/scope auth matrix against it.

jortel:
> 3\. **The one review question still open** ([thread](https://github.com/konveyor/tackle2-hub/pull/1119#discussion_r3777828975)): architect remains `get`-only on `agentic.agents` and `agentic.gateways`, so agent/gateway authoring is admin-only. Intended? The answer decides which personas get the management screens in the console.
Yes.

@djzager can you comment on this?

jortel:
@coderabbitai review

djzager:
Do I understand correctly that the `admin` in this scenario is saying that these resources can only be `kubectl create`d? For dev-preview that seems totally reasonable to me. My suspicion is that this pretty well aligns with what integrating with openshell would look like (admin only but architect could see them).

PR #1120, opened 2026-08-10
:bug: Fix hard coded path.

### LINKED ISSUE #1118: shared/nas.RmDir hardcodes /usr/bin/rm — fails on non-Linux platforms
## Summary

`shared/nas/dir.go` shells out to `/usr/bin/rm -rf` for directory removal:

```go
func RmDir(path string) (err error) {
    cmd := command.New("/usr/bin/rm")
    cmd.Options.Add("-rf", path)
    err = cmd.Run()
    return
}
```

This path doesn't exist on macOS (`/bin/rm`) and potentially other platforms. Since `RmDir` is called by `scm.Git.initHome()`, which runs at the start of every SCM method (`Fetch`, `Branch`, `Commit`, `Head`), the entire SCM package is not usable in mac.

## Impact

External consumers of `shared/scm` (e.g. the agentic-controller harness) can't run tests on macOS, and multi-arch support is limited.

jortel:
closes #1118 

Note: Using `rm` command instead of os.RemoveAll() because it's 12x faster.  Especially on network FS.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved directory copy and removal operations by allowing the system to locate the required commands dynamically.
  * Preserved existing command options and error handling.

### DIFF
--- a/shared/nas/dir.go
+++ b/shared/nas/dir.go
@@ -14,7 +14,7 @@ import (
 
 // CpDir copies (recursively) the directory.
 func CpDir(path, destination string) (err error) {
-	cmd := command.New("/usr/bin/cp")
+	cmd := command.New("cp")
 	cmd.Options.Add("-r", path, destination)
 	err = cmd.Run()
 	return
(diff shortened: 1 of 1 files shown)


aufi (approved):
LGTM

